PR #3, opened 2016-01-04
Fix Unit Tests

waigani:
Remove unit tests that no longer make sense. Skip two unit tests that
require a fake default .lingo to be written, add a todo to come back
to them.

Update failing tests so they pass.


guycook:
LGTM

PR #15, opened 2016-01-04
Integration testing

guycook:
@waigani this is the integration testing you would have seen already, it didn't seem to survive the bitbucket -> github migration


guycook:
CircleCI is choking on the config changes - closing and retrying...

PR #16, opened 2016-01-05
Integration testing

guycook:
The manual integration tests from BitBucket along with config to use CircleCI


guycook:
@waigani  As far as I can see the configuration is correct at this point, but the test times out due to building the docker/license tenet without any cache. Will need further investigation later.

PS. When I do have it sorted I'll squash all those commits together. Unfortunately each new test on Circle requires a new commit.

PR #19, opened 2016-01-05
Fixes for lints in tenet/tenet

guycook:


tenet/tenet.go
```diff
@@ -165,7 +165,8 @@ func (t *tenetService) Close() error {
 // 	}
 // }
 
-var KillAllTenetsErr = errors.New("kill all tenets")
+// TODO: Is this meant for production?
```
waigani: I think we can safely remove this - I can find no other reference to it.

PR #21, opened 2016-01-05
Make the binary driver use tcp connections on OS X. Fixes #20

### LINKED ISSUE #20: Fails to start / talk to tenets on OS X
This appears to be because OS X doesn't allow a net.ListenUnix call with a unix address with no Name.

Changing these to use a :0 bound TCP Listener appears to make them run properly, and accept the source.

I'm willing to work on a PR against this and the tenets repo to fix this, if there's interest.

AmandaCameron:
This relies on a change in the tenets package that I will be filing shortly, however this code is what made it start running properly for me on my OS X machine.


waigani:
Reviewed, checked out branches and tested manually. LGTM. Nice work!

PR #24, opened 2016-01-13
Render tenet docs to HTML

### LINKED ISSUE #23: "lingo docs" opens raw markdown in editor
The "lingo docs" command opens raw markdown, and at least on Linux, opens it in an editor (that's the behaviour of xdg-open). It would be nice if the markdown were rendered to HTML and opened in a browser.

axw:
"lingo docs" will now render the tenet docs
to HTML. Because the content comes from tenets,
and tenets could come from anywhere, we sanitise
the HTML output before presenting to the user.

Also changed "writeTenetDoc" to return errors,
logging in the "writeDoc" top-level command
function. In doing this, I found a data race:
multiple goroutines writing to the same map.
I've changed them to write to pre-sized slice
elements.

Fixes #23

### DIFF
--- a/commands/docs.go
+++ b/commands/docs.go
@@ -1,11 +1,13 @@
 package commands
 
 import (
+	"bytes"
 	"fmt"
 
 	"github.com/codegangsta/cli"
-	"github.com/lingo-reviews/tenets/go/dev/tenet/log"
+	"github.com/microcosm-cc/bluemonday"
 	"github.com/pkg/browser"
+	"github.com/russross/blackfriday"
 )
 
 var DocsCMD = cli.Command{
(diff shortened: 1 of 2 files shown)


waigani:
Nice! Looks like you committed a main binary by mistake? Also, the rendered HTML contained these funky characters: â€˜//â€™

They are in the source, not just the browser rendered text. I'm happy for this to land once the main is taken out. We can add an issue to chase down the UTF8 symbols - it'll be something in the dependencies.

axw:
I *removed* main :)

I'll look into the output issue.

axw:
@waigani PTAL. The blackfriday function I was using was enabling an extension that translates quotes to different HTML character entities depending on the context. Those were then being interpreted by bluemonday. I've just disabled the extensions for now.

waigani:
LGTM

PR #25, opened 2016-01-13
command: Remove diff logic

waigani:
Whether an issue is inside the diff context is calculated in the tenet. Remove
legacy code which calculated it in the client.


guycook:
LGTM

PR #29, opened 2016-01-27
Add tests for common public interface and fix some found bugs

guycook:


commands/common.go
```diff
@@ -194,15 +189,18 @@ func (c *config) RemoveTenet(name string, group string) error {
 		return errors.Trace(err)
 	}
 	var tenets []TenetConfig
-	for _, t := range c.AllTenets() {
+	err = errors.Errorf("tenet %q not found", name)
+	for _, t := range g.Tenets {
 		if t.Name != name {
 			tenets = append(tenets, t)
```
waigani: I think it would be clearer if we returned within the if (and set g.Tenets = tenets there). Then just return an error after the for loop. A few less lines of code that way.

commands/common_test.go
```diff
@@ -164,3 +165,412 @@ func (*CMDTest) TestMockContext(c *gc.C) {
...
+	c.Assert(complexConfig.HasTenetGroup("default"), gc.Equals, true)
+	c.Assert(complexConfig.HasTenetGroup("Group1"), gc.Equals, true)
+	c.Assert(complexConfig.HasTenetGroup("Group2"), gc.Equals, true)
+	c.Assert(complexConfig.HasTenetGroup("invalid"), gc.Equals, false)
+	c.Assert(complexConfig.HasTenetGroup("group1"), gc.Equals, false)
+}
+
+func (*CMDTest) TestAddTenetGroup(c *gc.C) {
+	// AddTenetGroup should add a new empty group to the config if that group doesn't already exist
+	testConfig := deepcopy.Iface(complexConfig).(*config)
```
waigani: No need for the deepcopy package:

```go
testConfig := &config{}
*testConfig = *complexConfig
```

commands/common_test.go
```diff
@@ -164,3 +165,412 @@ func (*CMDTest) TestMockContext(c *gc.C) {
...
+	err = testConfig3.AddTenet(TenetConfig{
+		Name:    "lingo-reviews/license",
+		Options: nil, // Test different options
+	}, "Group1")
+	c.Assert(err, gc.NotNil)
+}
+
+func (*CMDTest) TestRemoveTenet(c *gc.C) {
+	// RemoveTenet should remove a tenet only from the named group, and return an error if either the group or tenet is not found
+	testConfig1 := deepcopy.Iface(complexConfig).(*config)
```
waigani: Same deal here (and all other places), no need for deepcopy.

commands/common_test.go
```diff
@@ -164,3 +165,412 @@ func (*CMDTest) TestMockContext(c *gc.C) {
...
+func (*CMDTest) TestHasTenetGroup(c *gc.C) {
+	// HasTenetGroup should return true only for existing group names, case sensitive
+	c.Assert(complexConfig.HasTenetGroup("default"), gc.Equals, true)
+	c.Assert(complexConfig.HasTenetGroup("Group1"), gc.Equals, true)
+	c.Assert(complexConfig.HasTenetGroup("Group2"), gc.Equals, true)
+	c.Assert(complexConfig.HasTenetGroup("invalid"), gc.Equals, false)
+	c.Assert(complexConfig.HasTenetGroup("group1"), gc.Equals, false)
+}
+
+func (*CMDTest) TestAddTenetGroup(c *gc.C) {
```
waigani: It's great to see these unit tests coming in.

waigani:
Just some minor points, otherwise LGTM. Thank you :)

commands/common.go
guycook: @waigani I'm not sure if it's possible to be any terser here. Since this is a sort of remove-by-omission you have to let the loop complete, otherwise you'll get e.g. `[1, 2, 3, 4].Remove(3) -> [1, 2]` (If you look at the diff line 150 that's actually also the reason the `break` was removed - it was chopping the tail)

I could use a bool seenTenet instead which might communicate intent more clearly but would actually be a few lines longer

waigani: @guycook durr, of course - ignore me :) What you have is fine.

commands/common_test.go
guycook: I investigated this a bit: I think there's no way to avoid a deep copy (either by package or manually). For the Test*Group tests it's ok but eg. TestAddTenet mutates the config deeply at references which remain shared. I couldn't even resolve it by splitting out to seperate test functions as the one `complexConfig` is shared for the entire test run, and the function call order is undefined.

guycook:
@waigani If you're ok with the above comments can you merge this in so I can look at resolving the conflicts that will pop up with PR #31 this evening? Cheers

commands/common_test.go
waigani: Does it need to be a global? If not, make it a field on CMDTest and set it in SetupTest:

```go
(c *CMDTest) SetUpTest(c *gc.C) {
c.complexConfig = &config{...}
}
```

It'll be set before each unit test.

guycook:
No more deepcopy dependency, everything should be good to go

PR #31, opened 2016-01-28
Service Sub Commands

waigani:
This branch introduces sub commands to `lingo review` to send the result of a
review to a service. Configuration for each service is written in
$LINGO_HOME/services.yaml

The branch also introduces the `lingo copr` command which stands for "checkout
pull request". run `lingo copr --help` to see how it works.

A third command `lingo update` was added. This is to be run after the lingo
client tool has been upgraded. The command updates any required resources in
$LINGO_HOME.

As part of the refactoring, commands/common.go was moved into its
own package. The PR also contains several drive-by bug fixes and code clean-ups.


commands/copr.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/codegangsta/cli"
+	"github.com/lingo-reviews/lingo/util"
+)
+
+var CoprCMD = cli.Command{
+	Name:  "copr",
+	Usage: "Checkout a git pull request",
+	Description: `
+
+cd into a local git repository the pull request targets. The run:
```
guycook: This could probably use a rewrite, it's a bit unclear

commands/copr.go
```diff
@@ -0,0 +1,122 @@
...
+			Name:  "fetch-all,f",
+			Usage: "Fetch all remotes before checking out pull request. This will ensure the correct fork point can be found.",
+		},
+	},
+	Action: copr,
+}
+
+func copr(c *cli.Context) {
+	cliArgs := c.Args()
+	if l := len(cliArgs); l != 1 {
```
guycook: Could use our library function `common.ExactArgs` here

commands/copr.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	repoPath, err := exec.Command("git", "rev-parse", "--show-toplevel").Output()
+	if err != nil {
+		common.OSErrf(err.Error())
+		return
+	}
+	repo := path.Base(strings.Trim(string(repoPath), "\n"))
+	var fetchAll string
+	if c.Bool("fetch-all") {
```
guycook: Would it be nicer UI to have this on by default, otherwise user might be surprised by errors

util/util.go
```diff
@@ -183,3 +183,51 @@ func FormatOutput(in interface{}, tmplt string) (string, error) {
 
 	return out.String(), nil
 }
+
+func CheckoutPR() {
```
guycook: This doesn't ever seem to be called - additionally I'm a bit concerned about the duplication of the shell script below. Can it be unified with the one over in commands/copr.go

guycook:
Looks good besides the minor notes above. I would really like to have some integration tests for copr and at least one service (maybe just a dummy one?) because they're quite difficult to check manually. That might be a new issue for after next release though

PR #33, opened 2016-02-06
Service metrics

waigani:
Add TenetName, Discard and DiscardReason to issues. This will be used to track
tenet false positives.

### DIFF
--- a/commands/review/confirm.go
+++ b/commands/review/confirm.go
@@ -1,13 +1,15 @@
 package review
 
 import (
+	"bufio"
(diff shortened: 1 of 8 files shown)


guycook:
This all looks ok except it has the side effect of changing UI output in an unintuitive way: right now if you discard all issues you get `Done! Found 0 issues` as output - which this commit you will get `Done! Found 3 issues` even if you discard all.

@waigani Is this something you want to fix for this PR or should I just merge it and note the issue?

PR #37, opened 2016-02-08
Implement global settings api and find-all flag

guycook:
This shouldn't be merged until at least https://github.com/lingo-reviews/tenets/pull/39 has landed, and perhaps should be held off until we approach a new point release as it will require rebuilding of all tenets for the feature to be usable


waigani:
LGTM

PR #41, opened 2016-02-10
Binary Pull

waigani:
Lingo pull now also pulls source code from tenets and builds and installs the
binary.

Move the build code into its own package to be used by both pull and build commands.


guycook:
LGTM

PR #72, opened 2018-09-20
Added tenets for formatted strings and unsafe goroutines

Daanikus:


tenets/codelingo/go/unsafe-goroutine/codelingo.yaml
```diff
@@ -0,0 +1,33 @@
...
+      # Repeated code under go.for_stmt and go.range_stmt - waiting for macrofacts
+        any_of:
+          go.range_stmt(depth = any):
+            go.go_stmt(depth = any):
+              go.expr_stmt(depth = any):
+                @review.comment
+                go.ident:
+                  name as identName
+              exclude:
+                go.names(depth = any):
```
BlakeMScurr: I'm pretty sure this one shouldn't be at `(depth = any)`, or else you'll get a false negative for functions this, because the `unsafeItem` _does_ exist in a `go.names`. 
```
func uncaughtUnsafe() {
    for _, unsafeItem := range list {
        go func() {
            go func(unsafeItem int) {}(unsafeItem)
            // do something with item 
            unsafeItem
        }()
     }
}
```

BlakeMScurr: Also, what about variables that are declared in the go routine? Won't they give you a false positive?

```
func falsePositive() {
    for _, _ := range list {
        go func() {
            safeItem := 1
            // do something with item 
            safeItem
        }()
     }
}
```

Daanikus: The go.names(depth = any) was supposed to be removed, as it was from the block below. Fixing that and working on a way to prevent that false positive case

tenets/codelingo/go/string-format/codelingo.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - name: find-funcs
+    doc: Find formatted strings missing variable(s).
+    flows:
+      codelingo/review:
+        comment: This formatted string is missing variable(s)
+    query: |
+      import codelingo/ast/go
+      @review.comment
+      go.call_expr(depth = any):
+        go.selector_expr
```
BlakeMScurr: Is `go.selector_expr` necessary? It seems like it's here to express the assumption that calls to functions in external packages (like `fmt` or `errors`) will require arguments for each format verb, whereas calls in the same package won't. 

This is decent and maybe you should keep it, but at least add comment explaining the following:
This will get some false negatives - here's an example from the standard library where a call to a local function (`Sprintf`) should have an argument for each verb https://golang.org/src/fmt/print.go?s=5747:5797#L201. This example brings up another point, which is that you should account for variadic args https://gobyexample.com/variadic-functions. Also I think there could be false positives, for example, you could pass an SQL query with format arguments into a function and not expect the verbs to be resolved.

Another possible heuristic would be to only have this apply to functions ending in "f" like "Errorf" or "Sprintf". This is obviously not perfect, but it can be improved when we add more NLP.

tenets/codelingo/go/unsafe-goroutine/codelingo.yaml
BlakeMScurr: You should be able to resolve the false positive with another `exclude`.

tenets/codelingo/go/unused-private-functions/codelingo.yaml
```diff
@@ -0,0 +1,20 @@
...
+        @review.comment
+        go.ident:
+          private == "true"
+          name as funcName
+          exclude:
+            name == "main"
+      exclude:
+        go.call_expr(depth = any):
+          go.ident:
+            name as funcName
```
BlakeMScurr: This should be explicitly scoped to a directory with a `go.dir(depth = any)` fact. Reviews are scoped to the directory that `lingo run review` is run in, so for each private function this will search every subdir to find a call.

PR #73, opened 2018-09-26
add go global-var test

mullikine:


tenets/codelingo/go/global-var/codelingo.yaml
```diff
@@ -0,0 +1,18 @@
...
+      import codelingo/ast/go
+      
+      go.file(depth = any):
+        go.decls:
+          go.value_spec(depth = 1):
+            go.names:
+              @playground.highlight
+              go.ident:
+                private == "true"
+                public == "false"
```
BlakeMScurr: These two lines say the same thing, you can remove one.

BlakeMScurr: In fact, why not match public globals too?

tenets/codelingo/go/global-var/example.go
```diff
@@ -0,0 +1,3 @@
+package code
+
+var theVar = 2
```
BlakeMScurr: Might be good to have some test cases for non globals too.

tenets/codelingo/go/global-var/codelingo.yaml
waigani: Public globals may be used by another package

mullikine: Is this a reason for or against allowing public globals?

tenets/codelingo/go/global-var/example.go
```diff
@@ -0,0 +1,92 @@
...
+var myVar1, myVar2 = 1, 2
+
+var _ = 0
+
+// Those are not errors
+var myVar = 1
+
+// Those are fake errors which are currently not detected
+// because they start with 'err' or 'Err' and we don't
+// check if such a variable implements the error interface.
```
BlakeMScurr: I don't see anything in the tenet about checking whether the name of the ident starts with `err` or `Err`. Is this comment accurate?

mullikine: It may need to be split into 2 tenets to cover everything.

mullikine: No idea why those test cases were in there. They did come from the external tool though. I'll make sure I'm getting the right test cases.

mullikine: I see. The owner of the original linter must have re-used some test cases from a different linter. But they are actually testing global variables inside the 'error' code.

PR #74, opened 2018-09-27
noinit tenet

mullikine:


tenets/codelingo/go/init/expected.json
```diff
@@ -0,0 +1,46 @@
...
+    "end": {
+     "filename": "example.go",
+     "Offset": 61,
+     "Line": 7,
+     "Column": 16
+    }
+   },
+   "comment": "Init functions cause an import to have a side effects, and side effects are hard to test, reduce readability and increase the complexity of code.",
+   "ctxBefore": "type S struct{}\n",
+   "lineText": "func (s S) init(arg string) bool {",
```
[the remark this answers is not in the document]
waigani: Yep, so we should be searching for a go.func_decl in the global space, probably a direct child of go.decl.

tenets/codelingo/go/init/codelingo.yaml
```diff
@@ -0,0 +1,19 @@
...
+    query: |
+      import codelingo/ast/go
+      
+      go.file(depth = any):
+        go.decls:
+          go.func_decl:
+            @ review.comment
+            go.ident:
+              name == "init"
+            go.func_type:
```
waigani: This is an indirect way of finding the init func. You'll still hit a false positive for

```
func (r *typ) init() {
// ...
}
```

i.e. remove the field_list condition. Can we distinguish between methods and functions? Functions should not have a receiver.

mullikine: Yup I used that to check the arity of the function was == 0. Not positive if this is also the way to check for receivers

PR #75, opened 2018-09-28
Added tenet to find unused private functions in a package

Daanikus:
Tenet searches for functions that have not been used and comments on them


Daanikus:
Added expected results file

tenets/codelingo/go/unused-private-functions/codelingo.yaml
```diff
@@ -0,0 +1,20 @@
+tenets:
+  - name: find-unused-private-funcs
+    doc: Finds unused private functions.
+    flows:
+      codelingo/review:
+        comment: This private function is not used in the package.
+    query: |
+      import codelingo/ast/go
+
+      go.func_decl(depth = any):
```
BlakeMScurr: This should be scoped to a package fact like `go.package`.

tenets/codelingo/go/unused-private-functions/codelingo.yaml
```diff
@@ -0,0 +1,20 @@
...
+        @review.comment
+        go.ident:
+          private == "true"
+          name as funcName
+          exclude:
+            name == "main"
```
BlakeMScurr: We should also exclude the `init` function https://golang.org/doc/effective_go.html?#init

PR #77, opened 2018-09-30
new tenet: unconvert

mullikine:


tenets/codelingo/go/unconvert/codelingo.yaml
```diff
@@ -0,0 +1,20 @@
...
+      Identify unnecessary type conversions
+    flows:
+      codelingo/review:
+        comment: Unnecessary type conversion {{name}}; i.e., expression {{name}}(x) where x already has type {{name}}.
+    query: |
+      import codelingo/ast/go
+      go.file(depth = any):
+        go.call_expr(depth = any):
+          go.ident:
+            child_count == 0
```
waigani: Do you need the child_count == 0?

mullikine: It's probably not necessary. I left it in to be more explicit as I'm still little concerned my golang knowlege is limited.

PR #78, opened 2018-10-01
new tenet: unused private

mullikine:


tenets/codelingo/go/unused_private/codelingo.yaml
```diff
@@ -0,0 +1,30 @@
+tenets:
+  - name: unused_private
+    doc: |
+      Find unused private const, var, func, struct in Go.
```
waigani: Your example code doesn't test const or structs. Please add them.

tenets/codelingo/go/unused_private/codelingo.yaml
```diff
@@ -0,0 +1,30 @@
...
+        go.decls:
+          go.gen_decl:
+            go.value_spec:
+              go.names:
+                @ review.comment
+                go.ident(depth = any):
+                  @ review.vars.thingname
+                  name as privateVarName
+                  private == "true"
+      go.file(depth = any):
```
BlakeMScurr: Private variables could be used in different files in the same package. You'll want to add a `go.dir` fact at the top of the query and look for direct `go.file` children:

```
go.dir:
  go.file:
     # stuff
  go.file:
    # things
```

I just made a card about a caveat for this method https://trello.com/c/pAAywze9/381-review-with-top-level-dir. So it'd be good to either figure out a workaround or put the caveat in the tenet.

PR #80, opened 2018-10-02
new tenet: unused-private-const-var-same-package

mullikine:
I'm splitting up my previous pull request into 3 tenets, which I will later put into their own bundle.

### DIFF
--- a/tenets/codelingo/go/unused-private-const-var-same-package/codelingo.yaml
+++ b/tenets/codelingo/go/unused-private-const-var-same-package/codelingo.yaml
@@ -5,22 +5,19 @@ tenets:
     flows:
       codelingo/review:
         comment: |
-          '{{thingname}}' is private yet it is not referenced anywhere else in its package.
-          It should probably be removed.
+          Private global {{globalname}} is never used.  It should probably be removed.
     query: |
       import codelingo/ast/go
 
       go.dir:
         go.file(depth = any):
-          @ review.vars.myfname
-          filename
           go.decls:
             go.gen_decl:
               go.value_spec:
                 go.names:
                   @ review.comment
                   go.ident:
-                    @ review.vars.thingname
+                    @ review.vars.globalname
                     name as privateVarName
                     private == "true"
         exclude:
--- a/tenets/codelingo/go/unused-private-const-var-same-package/codelingo.yaml
+++ b/tenets/codelingo/go/unused-private-const-var-same-package/codelingo.yaml
@@ -1,0 +1,32 @@
+tenets:
+  - name: unused-private-const-var-same-package
+    doc: |
+      Find private const or var which are unused in a given package.
+    flows:
+      codelingo/review:
+        comment: |
+          '{{thingname}}' is private yet it is not referenced anywhere else in its package.
+          It should probably be removed.
+    query: |
+      import codelingo/ast/go
+
+      go.dir:
+        go.file(depth = any):
+          @ review.vars.myfname
+          filename
+          go.decls:
+            go.gen_decl:
+              go.value_spec:
+                go.names:
+                  @ review.comment
+                  go.ident:
+                    @ review.vars.thingname
+                    name as privateVarName
+                    private == "true"
+        exclude:
+          go.file(depth = any):
+            go.call_expr(depth = any):
+              go.args:
+                go.ident:
+                  name == privateVarName
+                  private == "true"
--- a/tenets/codelingo/go/unused-private-const-var-same-package/codelingo.yaml
+++ b/tenets/codelingo/go/unused-private-const-var-same-package/codelingo.yaml
@@ -22,8 +22,6 @@ tenets:
                     private == "true"
         exclude:
           go.file(depth = any):
-            go.call_expr(depth = any):
-              go.args:
-                go.ident:
-                  name == privateVarName
-                  private == "true"
+            go.ident(depth = any):
+              name == privateVarName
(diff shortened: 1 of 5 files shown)


tenets/codelingo/go/unused-private-const-var-same-package/codelingo.yaml
```diff
@@ -0,0 +1,32 @@
+tenets:
+  - name: unused-private-const-var-same-package
+    doc: |
+      Find private const or var which are unused in a given package.
+    flows:
+      codelingo/review:
+        comment: |
+          '{{thingname}}' is private yet it is not referenced anywhere else in its package.
```
BlakeMScurr: `thingname`?  😆

tenets/codelingo/go/unused-private-const-var-same-package/codelingo.yaml
```diff
@@ -0,0 +1,32 @@
...
+          It should probably be removed.
+    query: |
+      import codelingo/ast/go
+
+      go.dir:
+        go.file(depth = any):
+          @ review.vars.myfname
```
BlakeMScurr: Is `myfname` used anywhere?

tenets/codelingo/go/unused-private-const-var-same-package/codelingo.yaml
```diff
@@ -0,0 +1,32 @@
...
+                    @ review.vars.thingname
+                    name as privateVarName
+                    private == "true"
+        exclude:
+          go.file(depth = any):
+            go.call_expr(depth = any):
+              go.args:
+                go.ident:
+                  name == privateVarName
+                  private == "true"
```
BlakeMScurr: I think you can shorten it down. Does this work?

```
go.dir:
  go.value_spec(depth = any):
    go.names:
      @ review.comment
      go.ident:
        @ review.vars.thingname
        name as privateVarName
        private == "true"
  exclude:
    go.call_expr(depth = any):
      go.args:
        go.ident:
          name == privateVarName
          private == "true"
```

tenets/codelingo/go/unused-private-const-var-same-package/codelingo.yaml
mullikine: OK I'll change it. haha.

waigani: Yeah, you should use a more descriptive var name e.g. {{globalname}}. Also, comment should be "Private global {{globalname}} is never used".

tenets/codelingo/go/unused-private-const-var-same-package/codelingo.yaml
mullikine: This is geting slightly different behavior. It also appears to have matched one of the structs, which may have been a good thing if I were attempting to combine the 3 tenets, but I'm separating the logic.

tenets/codelingo/go/unused-private-const-var-same-package/codelingo.yaml
mullikine: Thanks. I will remove this.

tenets/codelingo/go/unused-private-const-var-same-package/codelingo.yaml
BlakeMScurr: Oh interesting. I guess that the `go.gen_decl` fact is what defines a variable/const definition. Can you at least remove the `go.file` facts without changing the output?

tenets/codelingo/go/unused-private-const-var-same-package/codelingo.yaml
```diff
@@ -0,0 +1,29 @@
...
+              go.value_spec:
+                go.names:
+                  @ review.comment
+                  go.ident:
+                    @ review.vars.globalname
+                    name as privateVarName
+                    private == "true"
+        exclude:
+          go.file(depth = any):
+            go.call_expr(depth = any):
```
BlakeMScurr: Does it matter that the var/const is used in a function call? If you changed the `foo` function to:
```
func foo() {
    a := privateVarA
    fmt.Printf("%s\n", a)
}
```

I suspect your tenet would suddenly think `privateVarA` was unused.

tenets/codelingo/go/unused-private-const-var-same-package/codelingo.yaml
BlakeMScurr: You can probably also remove the `go.decls` fact.

BlakeMScurr: (this isn't blocking btw, it's just nice to get them concise)

mullikine: I will give it a go.

mullikine: Sweet as. Getting insight as to which facts exactly are unique to a piece of syntax is what I'm hoping to achieve by making my CLQL a little verbose. Please bear with me :)

mullikine:
After simplifying the "exclude:" node to fix the issue of CLQL not matching a reference to an identifier which was not within a function call, there is no longer anything to differentiate it from the match above. I may need some advice on a workaround for this case.

mullikine:
This is waiting on a platform update to make it to release. I'll close this now and reopen when I can.

PR #83, opened 2018-10-03
Add parallel in tests tenet

emersonwood:


tenets/codelingo/jenkinsx/parallel-in-tests/codelingo.yaml
```diff
@@ -0,0 +1,49 @@
...
+        go.func_decl(depth = any):
+          go.ident:
+            name as funcName
+            regex(/Test.*/, funcName)
+          go.func_type:
+            go.field_list:
+              go.field:
+                go.names:
+                  go.ident:
+                    name == "t"
```
BlakeMScurr: Good point checking the reciever, I should add this to the integration test tenet. Ideally we'd test the type too.

PR #85, opened 2018-10-04
Integration test formatting

BlakeMScurr:
Check that integration tests match the correct format.

It's currently split into multiple cases, is verbose, and has caveats. This should eventually all be replaced by the ideal solution at the bottom of the file.


tenets/codelingo/jenkinsx/intformat/codelingo.yaml
```diff
@@ -0,0 +1,250 @@
...
+          # filename == append(packageName, "integration_test.go")
+          exclude:
+            # The only top level ident in a go file is (should be) the package name
+            go.ident:
+              name as testPackagename
+              regex(/_test/, testPackagename)
+              # name == append(packageName, "_test")
+
+## Better solutions:
+# Match files that have some characteristic of an integration test and are missing
```
waigani: I don't understand this sentence? Files that have a characteristic are missing a characteristic?

PR #86, opened 2018-10-04
new tenet for jenkins-x: orginterfaces

mullikine:


tenets/codelingo/jenkinsx/orginterfaces/codelingo.yaml
```diff
@@ -0,0 +1,21 @@
+functions:
+- name: isInterfaceFile
+  type: asserter
+  body: |
+    function (str) {
+      return ! str.match(/(^|\/)interface.go$/);
```
BlakeMScurr: Function should be called `isNotInterfaceFile`.  You could even create a more general function like this:

https://github.com/codelingo/codelingo/blob/master/tenets/codelingo/jenkinsx/intformat/codelingo.yaml#L8-L13

PR #92, opened 2018-10-10
Docs Flow

waigani:
Add a new Flow which generates contributor documentation from the Tenets of a
project.

### DIFF
--- a/flows/codelingo/docs/docs/docs.go
+++ b/flows/codelingo/docs/docs/docs.go
@@ -1,0 +1,84 @@
+package docs
+
+import (
+	"io/ioutil"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/juju/errors"
+)
+
+// Get all codelingo.yaml files that apply to the current review.
+func GetLingoFiles(workingDir string) (map[string][]byte, error) {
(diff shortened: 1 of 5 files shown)


flows/codelingo/docs/docs/docs.go
```diff
@@ -0,0 +1,84 @@
...
+
+	filename := filepath.Base(file)
+	return map[string]bool{
+		"codelingo.yaml": true,
+		"codelingo.yml":  true,
+	}[filename]
+}
+
+// by any git command-line actions
+func gitCMD(args ...string) (out string, err error) {
```
emersonwood: It'd be good if flows had a requirements / dependency section since there will be cases where people don't have git installed. Perhaps we could use docker containers for the flows instead of binaries?

flows/codelingo/docs/docs/parse/parse.go
```diff
@@ -0,0 +1,142 @@
+package parse
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+
+	"gopkg.in/yaml.v2"
+
+	"github.com/codelingo/clql/dotlingo"
```
emersonwood: Private dependency - may not be a biggy, but means people can't build from source.

PR #93, opened 2018-10-10
Fix Find Files

waigani:
Walk the file tree instead of using git to find codelingo.yaml files.


flows/codelingo/docs/codelingo.yaml
```diff
@@ -0,0 +1,10 @@
+tenets:
```
emersonwood: Not a proper tenet?

PR #99, opened 2018-10-11
Update ideal SQL injection tenet

BlakeMScurr:
Rewrite SQL injection tenet. 

Replace the magic features - `edge` is implemented, and `path` is well specced out.
Flesh out the details - it will actually work (with some specified caveats) one `path` syntax, and `"reaches"` and `"defines"` weave endpoints on the PHP lexicon are implemented.

### DIFF
--- a/tenets/codelingo/php/sql-concats/ideal-solution.lingo
+++ b/tenets/codelingo/php/sql-concats/ideal-solution.lingo
@@ -67,7 +67,7 @@ tenets:
       php.stmt_classmethod(depth = any):
         php.expr_assign:
           php.expr_methodcall(depth = any): # PHP surprisingly appears to have objects hanging off method calls
-            name == "get"
+            name == "get" #TODO: handle other http verbs - ideally more robustly
             php.expr_propertyfetch:
               name == "request"
 
--- a/tenets/codelingo/php/sql-concats/ideal-solution.lingo
+++ b/tenets/codelingo/php/sql-concats/ideal-solution.lingo
@@ -71,7 +71,8 @@ tenets:
             php.expr_propertyfetch:
               name == "request"
 
-           # Follow the reaches graph to function calls, but stop where the variable is passed into the sanitise function.
+          # Follow the reaches graph to function calls.
+          # Stop at variables defined by a call to the sanatise function.
           path(repeat = 0:):
             edge("reaches"):
               php.expr_assign:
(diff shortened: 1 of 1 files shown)


tenets/codelingo/php/sql-concats/ideal-solution.lingo
```diff
@@ -1,62 +1,127 @@
...
+        - Gets false positives if sanatise is run off the call path. In other words, we recurse along a
+          linear call chain from the handler to the db, if the input is sanatised by some other function
+          it will be considered unsafe where following "reaches".
+        - Doesn't catch direct calls to the database in the
+        - Possible reaches/defines issues:
+          - This tenet follows a direct reaching definition chain. This may be be confounded by having multiple
+            references to the same variable, especially across function boundaries.
+            In this example b->val is only indirectly given an unsafe value, and the $userInput definition
+            may not recursively reach $b->val:
+              <?php
```
waigani: pop src code in "```"

tenets/codelingo/php/sql-concats/ideal-solution.lingo
```diff
@@ -1,62 +1,127 @@
...
+                pathcontinue
+          pathend:
+            edge("defines"):
+              php.expr_variable:
+                id as callVar1
+        php.method_call(depth = any): # fork path to reach
+          php.expr_variable:
+            id == callVar1
+          # Recursively traverse to the function definition.
+          path(repeat = 0:):
```
waigani: What would it mean to repeat a path starting at >0?

waigani: It may be better to use a single integer for repeat and `-1` to indicate `forever`.

BlakeMScurr: Repeat is how many times you repeat the path. `(repeat = 0:)` means "repeat zero or more times". We could potentially expand it to `(repeat = 0:-1)` or `(repeat = 0:any)` if you think that's clearer - it's based on slice syntax right now, i.e., https://play.golang.org/p/4OXuSZNsc8J.

`(repeat = 1:)` is another important use case which means "repeat one or more times". I think it's important to have `(repeat = n:)` for n > 1 as well, because you could have a problem with 4 levels of nested if statements, or something. So it think it's important to be able to specify a range.

PR #101, opened 2018-10-12
Rewrite Flow

waigani:
Wire up the Rewrite Flow. Basic rewriting is functional: matched facts are
replaced with the string specified in the rewrite config. That string can
include rendered variables from the query.


sdk/flow/confirm.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+
+	// TODO(waigani) build up hunks here.
+
+	return true
+}
+
+func (c *hunkconfirmer) FormatPlainText(hunk *rewriterpc.Hunk) string {
+
+	xxx.Dump(hunk)
```
emersonwood: Remove debug code; this package can also bring in unnecessary dependencies.

PR #111, opened 2018-10-18
Rewrite Flow Refactor

waigani:
- Refactor to parse decorator args like cli input, extracting flags and args.
- Generalise the Flow SDK. Use the updated SDK to refactor the Rewrite Flow.
Other Flows should now be able to reuse the SDK.


sdk/flow/cli.go
```diff
@@ -3,38 +3,175 @@ package flow
...
+	}
+
+	// If user is manually confirming reviews, set a long timeout.
+	timeout := time.After(time.Hour * 1)
+
+	wg := sync.WaitGroup{}
+	wg.Add(1)
+	go func() {
+		wg.Wait()
+		close(decoratedResultc)
```
BlakeMScurr: Won't this panic, because the channel is already closed in a defer at the top of the function?

sdk/flow/cli.go
```diff
@@ -3,38 +3,175 @@ package flow
...
+			close(decoratedResultc)
+		}
+	}()
+
+	resultc, errc, cancel, err := f.RunCLI()
+	if err != nil {
+		return nil, errors.Trace(err)
+	}
+
+	// If user is manually confirming reviews, set a long timeout.
```
BlakeMScurr: Reference to "review" should be generic.

sdk/flow/cli.go
```diff
@@ -3,38 +3,175 @@ package flow
...
+l:
+	for {
+		select {
+		case err, ok := <-errc:
+			if !ok {
+				errc = nil
+				break
+			}
+
+			util.Logger.Debugf("Review error: %s", errors.ErrorStack(err))
```
BlakeMScurr: "review"

sdk/flow/cli.go
waigani: Only if we hit an error. A better pattern is to use a wait chan, I'll update it.

PR #116, opened 2018-10-18
Jx rewrite

mullikine:


mullikine:
These tenets have been rebased with master. This now requires testing.

PR #134, opened 2018-10-23
nil_only_functions fix: Don't trigger when nil and non nil are returned.

rstorr:


tenets/codelingo/go/nil_only_functions/codelingo.yaml
```diff
@@ -1,3 +1,10 @@
+functions:
```
waigani: new syntax is `funcs`

tenets/codelingo/go/nil_only_functions/expected.json
```diff
@@ -0,0 +1,68 @@
+[
+  {
+   "name": "nil-only-functions",
```
waigani: Are you using the latest review flow? The result format should be different.

tenets/codelingo/go/nil_only_functions/codelingo.yaml
rstorr: keep as functions until next deploy..

tenets/codelingo/go/nil_only_functions/expected.json
rstorr: will update.

PR #157, opened 2018-11-01
Declare empty slice

rstorr:


tenets/codelingo/go/declare_empty_slice/codelingo.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+      @review comment
+      go.assign_stmt(depth = any):
+        go.rhs:
+          go.composite_lit:
+            go.array_type:
+              go.ident
+
+              exclude:
+                go.basic_lit
```
BlakeMScurr: Do you need this? Shouldn't excluding `go.elts` be enough? Even a slice with a single element has a `go.elts` fact.

tenets/codelingo/go/declare_empty_slice/example.go
```diff
@@ -0,0 +1,18 @@
...
+
+  var t []string
+  t := []string{}
+  t := [2]string{2,3}
+  
+  i := 0
+  for i < 10 {
+    var w []string
+    w := []string{}
+    w := [2]string{2,3}
```
BlakeMScurr: I don't think this will compile, but that's a minor issue - it does parse.

rstorr: which line? the whole loop?

waigani: yep, needs to be type `int`

tenets/codelingo/go/declare_empty_slice/codelingo.yaml
rstorr: You're right I'll take it out.

PR #163, opened 2018-11-04
new k8 tenets: well-formed-lists well-named-interface well-named-lists

mullikine:
A few working k8 tenets

### DIFF
--- a/tenets/codelingo/k8/well-formed-lists/codelingo.yaml
+++ b/tenets/codelingo/k8/well-formed-lists/codelingo.yaml
@@ -10,29 +10,28 @@ tenets:
     query: |
       import codelingo/ast/go
       
-      go.file(depth = any):
-        go.decls:
-          go.gen_decl:
-            go.type_spec:
-              @review comment
-              go.ident:
-                name as listName
-              go.struct_type:
-                go.field_list:
+      go.decls:
+        go.gen_decl:
+          go.type_spec:
+            @review comment
+            go.ident:
+              name as listName
+            go.struct_type:
+              go.field_list:
+                go.field:
+                  go.selector_expr:
+                    go.ident:
+                      name == "metav1"
+                    go.ident:
+                      name == "TypeMeta"
+                go.field:
+                  go.selector_expr:
+                    go.ident:
+                      name == "metav1"
+                    go.ident:
+                      name == "ListMeta"
+                exclude:
                   go.field:
-                    go.selector_expr:
+                    go.names:
                       go.ident:
-                        name == "metav1"
-                      go.ident:
-                        name == "TypeMeta"
-                  go.field:
-                    go.selector_expr:
-                      go.ident:
-                        name == "metav1"
-                      go.ident:
-                        name == "ListMeta"
-                  exclude:
-                    go.field:
-                      go.names:
-                        go.ident:
-                          name == "Items"
+                        name == "Items"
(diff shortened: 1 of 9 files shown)


tenets/codelingo/k8/well-formed-lists/codelingo.yaml
```diff
@@ -0,0 +1,38 @@
...
+      codelingo/docs:
+        title: Well Formed Lists
+        body: |
+          All lists use the required `items` field to contain the array of objects they return.
+      codelingo/review:
+        comment: This is a list and must contain an `items` field.
+    query: |
+      import codelingo/ast/go
+      
+      go.file(depth = any):
```
rstorr: Don't think you need to specify go.file

tenets/codelingo/k8/well-named-interface/codelingo.yaml
```diff
@@ -0,0 +1,30 @@
...
+      codelingo/docs:
+        title: Well Named Interface
+        body: |
+          Please consider the package name when selecting an interface name, and avoid redundancy. e.g. `storage.Interface` is better than `storage.StorageInterface`.
+      codelingo/review:
+        comment: Interface name {{interfaceName}} has name of package {{packageName} redundantly prefixed.
+    query: |
+      import codelingo/ast/go
+            
+      go.file(depth = any):
```
rstorr: Don't think you need to specify go.file

tenets/codelingo/k8/well-named-interface/codelingo.yaml
```diff
@@ -0,0 +1,30 @@
...
+        go.ident:
+          name as packageName
+        go.decls:
+          go.gen_decl: # this can probably be made depth = any
```
rstorr: I think an interface is only going to be at the top level so no need for the comment. Although correct me if I'm wrong.

tenets/codelingo/k8/well-named-lists/codelingo.yaml
```diff
@@ -0,0 +1,41 @@
...
+      codelingo/docs:
+        title: Well Named Lists
+        body: |
+          The name of a list kind must end with List.
+      codelingo/review:
+        comment: The kind `{{listName}}` is a collection of resources and its name must end in `List`.
+    query: |
+      import codelingo/ast/go
+      
+      go.file(depth = any):
```
rstorr: no need for go.file

tenets/codelingo/k8/well-named-interface/codelingo.yaml
mullikine: This one needs go.file: because when followed by go.ident: it is a package name

tenets/codelingo/k8/well-named-interface/codelingo.yaml
BlakeMScurr: You can define interfaces anywhere, we do need `depth = any`. https://play.golang.org/p/r-UyQhKljtQ

tenets/codelingo/k8/well-named-interface/codelingo.yaml
mullikine: That being said, it's only going to match when there is an interface_type. As you mentioned next, an interface is only going to be at the top level, so technically, this may not need the go.file: due of the interface which follows. However, I feel like removing this go.file: would make the tenet non-migratable if in future there was a package definition fact and relying on side-effects that may change in future seems nasty.

tenets/codelingo/k8/well-named-interface/codelingo.yaml
mullikine: Actually, I don't know for sure if the interface is always going to be at the top level. I imagine it's possible to have them defined elsewhere. Unsure about this.

Issue #177, opened 2018-11-07
Can't use != on vars 


rstorr:
My workaround is to use a notEquals func as I can't currently do
name != bname

rstorr:
rstorr hit a uncaught error: children of excluded validator should not fill up excluded variable values

PR #178, opened 2018-11-07
fix well named lock tenet.

rstorr:


rstorr:
The error is that you can't currently != vars e.g

go.ident:
name != varname

my workaround was to use a notEquals func.

I filed an issue: https://github.com/codelingo/codelingo/issues/177

tenets/codelingo/kubernetes/well-named-lock/codelingo.yaml
```diff
@@ -6,32 +13,33 @@ tenets:
         body: |
           A lock should be called `lock` unless other locks are present.
       codelingo/review:
-        comment: There are no other locks besides '{{lockname}}' here. Rename '{{lockname}}' to 'lock'.
+        comment: A lock should be called `lock` unless other locks are present.
```
mullikine: See if you can keep this comment

PR #179, opened 2018-11-07
Fix shadowed func param

rstorr:


tenets/codelingo/kubernetes/shadowed-func-parameter/codelingo.yaml
```diff
@@ -4,33 +4,27 @@ tenets:
...
-            go.assign_stmt(depth = any):
-              go.lhs:
-                @rewrite "inner{{newname}}"
-                @review comment
+      go.func_decl(depth = any):
+        go.func_type:
+          go.field_list: 
+            go.field:
+              go.names:
                 go.ident:
```
mullikine: Shift the @review comment here to bring the number of matches down to 2. It's currently matching all inner idents but could be simplified by matching only the parameter. Tested and the tenet still works. LGTM otherwise!

PR #183, opened 2018-11-08
tenet working

rstorr:


rstorr:
Tenet generated from bullet point 2: https://gist.github.com/lavalamp/4bd23295a9f32706a48f

Unable to detect when interface is defined outside file though such as error: https://golang.org/doc/faq#nil_error

tenets/codelingo/kubernetes/misused-nil-interface/codelingo.yaml
```diff
@@ -13,69 +19,30 @@ tenets:
...
-          go.func_type
+            name as interfaceName
+            type as interfaceType
+          go.interface_type
         go.func_decl:
-          go.ident:
-            name as factoryfuncname
           go.func_type:
-            go.field_list
             go.field_list:
```
mullikine: Change to

```
          go.func_type:
            go.field_list
            go.field_list:
```

Otherwise it will also match any function with a parameter called interfaceType, not just return parameters

PR #184, opened 2018-11-09
fix tenet registry for lnd

mullikine:


registry/tenets.yaml
```diff
@@ -69,6 +69,16 @@ codelingo:
       - kubernetes
       - golang
       - go
+  lightning-network-daemon:
+    default:
```
waigani: s/default/lighting...

PR #187, opened 2018-11-09
New lnd tenet exported package is tested

mullikine:


mullikine:
packages are tested in _test.go files. we use go.dir because the packages being tested are exported publicly and may be in any dir, not always the same dir as the package

mullikine:
see section 4.2 https://github.com/lightningnetwork/lnd/blob/master/docs/code_contribution_guidelines.md

mullikine:
added to registry
removed extraneous clql file

PR #191, opened 2018-11-11
Added 3 Effective Go tenets

Daanikus:
Added tenets to check for:
Renamed imports
Subject is first word of comment
Underscored in names

### DIFF
--- a/tenets/codelingo/effective-go/avoid-renaming-imports/codelingo.yaml
+++ b/tenets/codelingo/effective-go/avoid-renaming-imports/codelingo.yaml
@@ -1,0 +1,33 @@
+funcs:
+  - name: isInvalid
+    type: asserter
+    body: |
+      function(name) {
+        return name !== "_"
+      }
+tenets:
+  - name: avoid-renaming-imports
+    flows:
+      codelingo/docs:
+        title: Avoid Renaming Imports
+        body: |
+          Avoid renaming imports except to avoid a name collision; good package names
+          should not require renaming. In the event of collision, prefer to rename the
+          most local or project-specific import.
+      codelingo/review:
+        comment: |
+          Avoid renaming imports except to avoid a name collision; good package names
+          should not require renaming. In the event of collision, prefer to rename the
+          most local or project-specific import.
+    query: |
+      import codelingo/ast/go
+
+      go.gen_decl(depth = any):
+        @review comment
+        go.import_spec:
+          go.ident:
+            name as impName
+            isInvalid(impName)
+          go.basic_lit
+
+      
--- a/tenets/codelingo/effective-go/avoid-renaming-imports/codelingo.yaml
+++ b/tenets/codelingo/effective-go/avoid-renaming-imports/codelingo.yaml
@@ -1,10 +1,3 @@
-funcs:
-  - name: isInvalid
-    type: asserter
-    body: |
-      function(name) {
-        return name !== "_"
-      }
 tenets:
   - name: avoid-renaming-imports
     flows:
@@ -26,8 +19,6 @@ tenets:
         @review comment
         go.import_spec:
           go.ident:
-            name as impName
-            isInvalid(impName)
-          go.basic_lit
+            name != "_"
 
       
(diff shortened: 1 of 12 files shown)


tenets/codelingo/effective-go/avoid-renaming-imports/codelingo.yaml
```diff
@@ -0,0 +1,33 @@
...
+    query: |
+      import codelingo/ast/go
+
+      go.gen_decl(depth = any):
+        @review comment
+        go.import_spec:
+          go.ident:
+            name as impName
+            isInvalid(impName)
+          go.basic_lit
```
rstorr: Is there going to be a case where an import spec doesn't have a basic_lit? If not remove it.

tenets/codelingo/effective-go/avoid-renaming-imports/codelingo.yaml
```diff
@@ -0,0 +1,33 @@
...
+          should not require renaming. In the event of collision, prefer to rename the
+          most local or project-specific import.
+    query: |
+      import codelingo/ast/go
+
+      go.gen_decl(depth = any):
+        @review comment
+        go.import_spec:
+          go.ident:
+            name as impName
```
waigani: Why not just `name != "_"` ?

tenets/codelingo/effective-go/comment-first-word-as-subject/codelingo.yaml
```diff
@@ -0,0 +1,32 @@
...
+  - name: comment-first-word-as-subject
+    flows:
+      codelingo/docs:
+        title: Comment First Word as Subject
+        body: |
+          Doc comments work best as complete sentences, which allow a wide variety of automated presentations. 
+          The first sentence should be a one-sentence summary that starts with the name being declared.
+      codelingo/review:
+        comment: |
+          The first sentence should be a one-sentence summary that starts with the name ({{funcName}}) being declared.
```
rstorr: We've been using `` for var names instead of brackets. e.g ...starts with func name `baz` being declared.

tenets/codelingo/effective-go/avoid-renaming-imports/codelingo.yaml
waigani: Also, you're not covering any named imports other than "_"? You should be able to check the import strings - if they both end in the same name, then a rename is okay, otherwise it's an error.

You should also be able to work out which is the most project-specific import.

tenets/codelingo/effective-go/underscores-in-name/codelingo.yaml
```diff
@@ -0,0 +1,22 @@
...
+    body: |
+      function(name) {
+        return name.indexOf("_") !== -1
+      }
+tenets:
+  - name: underscores-in-name
+    flows:
+      codelingo/docs:
+        title: Underscores in Name
+        body: The convention in Go is to use MixedCaps or mixedCaps rather than underscores to write multiword names.
```
rstorr: change to .... is to use PascalCase or camelCase... for docs and review comment

tenets/codelingo/effective-go/avoid-renaming-imports/codelingo.yaml
Daanikus: Will change that. I think there was a bug with the inequality operator when I wrote this. The query will only pick up imports that are named. Unnamed imports are ignored. The idea is to match any named import except for "_"

rstorr: @waigani is referring to an instance such as 

import (
    "text/template"
    htemplate "html/template" // this is now imported as htemplate
)

here it is needed that one of the template imports are named since they are both importing 'template' . So if more than one import has the same ending i.e 'template' then you can ignore it in the query.

Daanikus: Ah yup. I'll include that in the tenet

PR #196, opened 2018-11-12
Added EG tenet to check for good package name

Daanikus:
Tenet for https://golang.org/doc/effective_go.html#package-names


mullikine:
Add a readme like this. it provides a way to load the tenet in the playground

https://github.com/mullikine/codelingo/blob/lightning-network-daemon/tenets/codelingo/go/init/README.md

PR #197, opened 2018-11-12
Added EG tenet to enforce comment first word is subject

Daanikus:
Tenet for https://golang.org/doc/effective_go.html#commentary


mullikine:
LGTM, except the expected.json needs updating. https://ptpb.pw/4wyY.json

Daanikus:
JSON updated

PR #201, opened 2018-11-13
add status to tenets.

rstorr:


registry/tenets.yaml
```diff
@@ -18,31 +24,49 @@ codelingo:
...
+        status: complete
+      println-format-strings:
+        status: complete
+      reallocated-slice:
+        status: complete
+      sprintf:
+        status: complete
+      tested:
+        status: complete
+      undo:
```
mullikine: is this meant to be todo?

rstorr: thanks

PR #209, opened 2018-11-15
Added CRC tenet to check for use of math/rand in key creation (hasIssues)

Daanikus:


tenets/codelingo/code-review-comments/use-crypto-rand/codelingo.yaml
```diff
@@ -0,0 +1,39 @@
...
+          throwaway ones. Unseeded, the generator is completely predictable. 
+          Seeded with time.Nanoseconds(), there are just a few bits of entropy. 
+          Instead, use crypto/rand's Reader, and if you need text, print to 
+          hexadecimal or base64
+    query: |
+      import codelingo/ast/go
+
+      go.import_spec(depth = any):     
+        go.basic_lit:
+          value == "math/rand"
```
waigani: @Daanikus catch the case where the import is renamed:

```go
import myRand "math/rand"
```

 Add a test case for it.

tenets/codelingo/code-review-comments/use-crypto-rand/example.go
```diff
@@ -0,0 +1,23 @@
+package main
```
waigani: Why do you have two files exactly the same?

tenets/codelingo/code-review-comments/use-crypto-rand/expected.json
```diff
@@ -0,0 +1,8 @@
+[
+  {
+   "Comment": "Do not use package math/rand to generate keys, even \nthrowaway ones. Unseeded, the generator is completely predictable. \nSeeded with time.Nanoseconds(), there are just a few bits of entropy. \nInstead, use crypto/rand's Reader, and if you need text, print to \nhexadecimal or base64\n",
+   "Filename": "effective-go/use-crypto-rand/example-bad.go",
+   "Line": 4,
```
waigani: attach the comment to where it's being used to gen the key. There should be two results here.

tenets/codelingo/code-review-comments/use-crypto-rand/codelingo.yaml
```diff
@@ -0,0 +1,39 @@
...
+      any_of:
+        @review comment
```
mullikine: I remember you were having trouble with multiple review comments inside of an any_of. This seems to be not complaining about having multiple review comments anymore, which is good. However, it looks like you're getting multiple hits for the same thing. I assume this is because you have two review comments. Is it possible to move the review comment to a 'pipeline' such as any_of.. A question for  Blake

```
Running review flow...
example-bad.go:19:2

    Do not use package math/rand to generate keys, even
    throwaway ones. Unseeded, the generator is completely predictable.
    Seeded with time.Nanoseconds(), there are just a few bits of entropy.
    Instead, use crypto/rand's Reader, and if you need text, print to
    hexadecimal or base64


    ...
    )

  + func Key() string {
  +     buf := make([]byte, 16)
  +     _, err := rand.Read(buf)
  +     if err != nil {
  +             panic(err) // out of randomness, should never happen
  +     }
  +     return fmt.Sprintf("%x", buf)
  +     // or hex.EncodeToString(buf)
  +     // or base64.StdEncoding.EncodeToString(buf)
  + }

    ...

[o]pen [k]eep [d]iscard : k
example.go:19:2

    Do not use package math/rand to generate keys, even
    throwaway ones. Unseeded, the generator is completely predictable.
    Seeded with time.Nanoseconds(), there are just a few bits of entropy.
    Instead, use crypto/rand's Reader, and if you need text, print to
    hexadecimal or base64


    ...
    )

  + func Key() string {
  +     buf := make([]byte, 16)
  +     _, err := rand.Read(buf)
  +     if err != nil {
  +             panic(err) // out of randomness, should never happen
  +     }
  +     return fmt.Sprintf("%x", buf)
  +     // or hex.EncodeToString(buf)
  +     // or base64.StdEncoding.EncodeToString(buf)
  + }

    func main() {
    ...

[o]pen [k]eep [d]iscard : k
```

mullikine:
lingo appears to be hung after the last k

```
sudo strace -p "$(pgrep -a lingo | grep /bin/lingo | s field 1)"
strace: Process 24922 attached
waitid(P_PID, 24932,
```

tenets/codelingo/code-review-comments/use-crypto-rand/codelingo.yaml
```diff
@@ -0,0 +1,39 @@
+funcs:
+  - name: isKeyName
+    type: asserter
+    body: |
+      function(name) {
+        return name.toLowerCase().indexOf("key") !== -1
+      }
+tenets:
+  - name: use-crypto-rand
+    flows:
```
mullikine: Missing a docs section.

PR #210, opened 2018-11-16
Added CRC tenet to check for renamed imports

Daanikus:


tenets/codelingo/code-review-comments/avoid-renaming-imports/codelingo.yaml
```diff
@@ -0,0 +1,40 @@
...
+        go.import_spec(depth = 0:2):
+          go.ident: # Ident is the rename and absent if not named
+            name != "_"
+          go.basic_lit: # Basic_lit is the "import/path"
+            value as importOne
+        go.import_spec:
+          go.basic_lit:
+            value as importTwo
+        diffNameImport(importOne, importTwo)
+
```
mullikine: trim trailing whitespace

tenets/codelingo/code-review-comments/avoid-renaming-imports/codelingo.yaml
```diff
@@ -0,0 +1,40 @@
...
+    flows:
+      codelingo/docs:
+        title: Avoid Renaming Imports
+        body: |
+          Avoid renaming imports except to avoid a name collision; good package names
+          should not require renaming. In the event of collision, prefer to rename the
+          most local or project-specific import.
+      codelingo/review:
+        comment: |
+          Avoid renaming imports except to avoid a name collision; good package names
```
mullikine: This is the tense comments should be written in. 

```
This import renames the default identifier `{{originalImportName}}` to `{{newImportName}}` but this is a bad practice. Remove `{{newImportName}}` to keep the original identifier.
```

Or maybe rename to this
```
This import is renamed to `{{badIdentName}}` but this is a bad practice. Import without specifying a new name.
```

tenets/codelingo/code-review-comments/avoid-renaming-imports/codelingo.yaml
```diff
@@ -0,0 +1,40 @@
+funcs:
+  - name: diffNameImport
```
mullikine: Nitpicking but renaming this function to something like importNamesAreDifferent could prevent needing to add a comment here. Usually these asserter functions would be named starting with 'is' to make its function a little clearer but this has two parameters.

tenets/codelingo/code-review-comments/avoid-renaming-imports/expected.json
```diff
@@ -0,0 +1,32 @@
+[
```
mullikine: Getting duplicate hits. Please link to this in your tenet card.

https://github.com/mullikine/mullikine.github.io/blob/master/codelingo/troubleshooting/tenets/avoid-renaming-imports_multiple-review-hits.org

PR #215, opened 2018-11-16
Added EG tenet to check for package comments

Daanikus:


tenets/codelingo/effective-go/package-comment/codelingo.yaml
```diff
@@ -0,0 +1,17 @@
...
+        comment: |
+          Every package should have a package comment, a block comment preceding the package clause.
+          For multi-file packages, the package comment only needs to be present in one file, and any one will do.
+    query: |
+      import codelingo/ast/go
+
+      @review comment
+      go.file(depth = any):
+        exclude:
+          go.comment(depth = any):
```
waigani: Can you handle multi-file packages? If not, load up a blocked card and add an issue to the Tenet.

tenets/codelingo/effective-go/package-comment/codelingo.yaml
```diff
@@ -0,0 +1,17 @@
+tenets:
+  - name: package-comment
+    flows:
+      codelingo/review:
```
mullikine: need a docs flow here

tenets/codelingo/effective-go/package-comment/codelingo.yaml
mullikine: Give this a go

```
import codelingo/ast/go

@review comment
go.dir(depth = any):
  go.file:
    go.ident:
      name as packageName
    exclude:
      go.comment(depth = any):
        start_offset == 0
  exclude:
    go.file:
      go.ident:
        name == packageName
      go.comment(depth = any):
        start_offset == 0
```

tenets/codelingo/effective-go/package-comment/codelingo.yaml
```diff
@@ -0,0 +1,25 @@
...
+          the detailed documentation that follows.
+      codelingo/review:
+        comment: |
+          Every package should have a package comment, a block comment preceding the package clause.
+          For multi-file packages, the package comment only needs to be present in one file, and any one will do.
+    query: |
+      import codelingo/ast/go
+
+      @review comment
+      go.file(depth = any):
```
mullikine: im a little concerned about removing the go.dir. if it's removed then you'll have to add a check to make sure that the files being compared live in the same repository. by putting (depth = any) onto the file fact you allow it to compare any two files. this might be the cause of so many hits.

alternatively, you might be able to add a js function to compare the directories by using the filename fact from each file

BlakeMScurr: I'm not sure what you mean here. How are we comparing two files? There's only a single file fact in the query.

mullikine: Just saw this sorry. The 2nd `go.file` is on line 26

tenets/codelingo/effective-go/package-comment/codelingo.yaml
```diff
@@ -0,0 +1,32 @@
...
+        go.ident:
+          name as packageName
+        exclude:
+          go.comment(depth = any):
+            start_offset == 0
+      exclude:
+        go.file:
```
mullikine: Just noticed that the first go.file has (depth = any) but this go.file does not have a depth specified. I think we need to assert that both `go.file`s have the same depth. The best way to do that is probably to remove the (depth = any) from the go.file and put a parent go.dir(depth = any). Leaving it as it is will probably make the query as a whole match every file, which is bad and adding (depth = any) to this go.file is probably going to make compare all the files with each other which might be bad....

PR #221, opened 2018-11-19
Improve doc comment and review comment

Daanikus:


tenets/codelingo/go/global-var/codelingo.yaml
```diff
@@ -4,9 +4,11 @@ tenets:
       codelingo/docs:
         title: Global Variable
         body: |
-          Check that no globals are present in Go code.
+          Global variables are an input to functions that is not visible in 
+          the functions signature, complicate testing, reduces readability 
+          and increase the complexity of code.
       codelingo/review:
-        comment: Global variables are an input to functions that is not visible in the functions signature, complicate testing, reduces readability and increase the complexity of code.
+        comment: Avoid global variables to improve readability and reduce complexity
```
mullikine: This is much better, but I'd also put something along this lines of 'This is a global variable, avoid etc.'.

PR #223, opened 2018-11-20
Fix context-first-arg tenet

Daanikus:
Utilised sibling order of field lists to ensure we only look at the input parameters


tenets/codelingo/code-review-comments/context-first-arg/codelingo.yaml
```diff
@@ -17,10 +17,10 @@ tenets:
     query: |
       import codelingo/ast/go
       
-      # TODO use macro fact to ensure this is not a return parameter
       go.field_list(depth = any):
-        go.field
+        sibling_order == 0
```
mullikine: I also prefer this to having two go.field facts. Nice!
We also need to be adding workarounds to this, no matter how trivial. Looks like a good one to add to kick us off.
https://trello.com/c/yMxdjPVG/1888-add-workarounds-for-tenets-to-discusscodelingoio

PR #232, opened 2018-11-21
Add Cockroach tenet to check for inline comments on func args

Daanikus:
… call

### DIFF
--- a/tenets/codelingo/cockroachdb/func-args-inline-comments/codelingo.yaml
+++ b/tenets/codelingo/cockroachdb/func-args-inline-comments/codelingo.yaml
@@ -2,7 +2,7 @@ tenets:
   - name: func-args-inline-comments
     flows:
       codelingo/docs:
-        title: Context as First Argument
+        title: Function Arguments Inline Comments
         body: |
           A code reader encountering a function call should be able to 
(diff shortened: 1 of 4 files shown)


tenets/codelingo/cockroachdb/func-args-inline-comments/codelingo.yaml
```diff
@@ -0,0 +1,37 @@
...
+      # This tenet is not working
+      # Need to be able to query 'line'
+
+      go.file(depth = any):
+        go.call_expr(depth = any):
+          go.args:
+            @review comment
+            go.ident(depth = any):
+              name == "nil"
+              start_line == line
```
waigani: why not use start_offset? Also, one of them needs to be `as` not `==`.

Daanikus: I could try using offset although we won't know whether the comment is inline or not. This may not be an issue really but then we run into the other issue of having to compare offsets, which is currently not working

waigani: I've got a PR up to fix that: https://github.com/codelingo/clql/pull/90

PR #234, opened 2018-11-21
Add cockroach mock tenet for wrapping func signatures

Daanikus:


mullikine:
I think we're still adding mocks to the registry and the tenet_bundle.yaml. And according to @rstorr they go in with hasIssues = false. Might want to double check.

PR #239, opened 2018-11-22
Add cockroach tenet to check for correct verb use in calls to fmt

Daanikus:


tenets/codelingo/cockroachdb/fmt-verbs/codelingo.yaml
```diff
@@ -0,0 +1,89 @@
...
+        body: |
+          Prefer the most specific verb for your use. In other words, 
+          prefer to avoid %v when possible. However, %v is to be used 
+          when formatting bindings which might be nil and which do not 
+          already handle nil formatting. Notably, nil errors formatted 
+          as %s will render as "%!s()" while nil errors formatted as %v 
+          will render as "". Therefore, prefer %v when formatting errors 
+          which are not known to be non-nil.
+      codelingo/review:
+        comment:  type should use an appropriate verb such as `{{result}}` for `{{argType}}`
```
mullikine: getting MISSING for {{result}}

https://github.com/mullikine/mullikine.github.io/blob/master/codelingo/troubleshooting/tenets/fmt-verbs_getting-missing-for-comment-var.org

Daanikus: Yeah, forgot about that MISSING thing. It's a go thing so somewhere along the way the "%d" string is being interpreted as a verb. I'll look into a way to force it to be treated as a plain string

PR #249, opened 2018-11-26
Add CRC tenet to check for stuttering named result parameters

Daanikus:


mullikine:
I think this is just missing a readme.

Issue #212, opened 2018-11-16
Tenet: avoid-renaming-imports. Can't implement logic to compare only certain imports


Daanikus:
We'd like the tenet to exclude named imports if there is a naming conflict, but the tenet will compare each import to all other imports. This results in false positives.

BlakeMScurr:
fixed in https://github.com/codelingo/codelingo/pull/265

PR #265, opened 2018-11-29
Fix avoid-renaming-imports.

BlakeMScurr:
If there is a renamed import, assert that there is no other import with the same ending, rather than asserting that there is an import without the same ending. Make sure the two import statements are not the same.

### DIFF
--- a/tenets/codelingo/code-review-comments/avoid-renaming-imports/codelingo.yaml
+++ b/tenets/codelingo/code-review-comments/avoid-renaming-imports/codelingo.yaml
@@ -1,11 +1,17 @@
 funcs:
-  - name: importNamesAreDifferent
+  - name: importNamesEqual
     type: asserter
     body: |
       function(a, b) {
         var aSplit = a.toLowerCase().split("/")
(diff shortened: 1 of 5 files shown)


mullikine:
Cool. Looks good. Can this be closed now? https://github.com/codelingo/codelingo/issues/212 Also, just a reminder to remove the issues.yaml file from the repo and from the registry

tenets/codelingo/code-review-comments/avoid-renaming-imports/codelingo.yaml
```diff
@@ -1,11 +1,17 @@
...
+  - name: importNamesEqual
     type: asserter
     body: |
       function(a, b) {
         var aSplit = a.toLowerCase().split("/")
         var bSplit = b.toLowerCase().split("/")
-        return a[a.length - 1] !== b[b.length - 1]
+        return a[a.length - 1] === b[b.length - 1]
+      }
+  - name: neq
```
waigani: It would be more natural to have this as eq and then negate it in the query if CLQL supports that.

eq would then make a good generic function to import.

PR #271, opened 2018-12-06
Added rewrite to go-error-fmt

mullikine:


tenets/codelingo/code-review-comments/go-error-fmt/codelingo.yaml
```diff
@@ -7,9 +7,28 @@ funcs:
         var cap = a.charAt(0) >= "A" && a.charAt(0) <= "Z"
         return dot || cap
       }
+  - name: fixErrorFormat
+    type: resolver
+    body: |
+      function(text) {
+        function escapeGoFormat(text) {
+          return text.replace(new RegExp("([\\\"])", "g"), "\\$1")
```
BlakeMScurr: Are the square brackets needed? Doesn't that mean that the chars don't have to be in order?

Are there other chars that we might have to escape? Such as backtick?

mullikine: The two chars that the function is escaping are \ and ". The aim of this function is to prepend another \ onto every \ and ". There might be other chars.

PR #272, opened 2018-12-06
User variables

BlakeMScurr:
Stream user variable setters from the server. They are currently just set to default in the SDK. A further PR will build send a variable setter channel to be handled by each individual flow.

Relies on: https://github.com/codelingo/rpc/pull/10

### DIFF
--- a/flows/codelingo/review/review/review.go
+++ b/flows/codelingo/review/review/review.go
@@ -9,11 +9,10 @@ import (
 
 	"github.com/golang/protobuf/ptypes"
 
+	flowutil "github.com/codelingo/codelingo/sdk/flow"
 	"github.com/codelingo/lingo/app/util"
-	"github.com/codelingo/lingo/service"
 	grpcclient "github.com/codelingo/lingo/service/grpc"
 	"github.com/codelingo/rpc/flow"
-	"github.com/codelingo/rpc/flow/client"
 	"github.com/golang/protobuf/proto"
 	"github.com/juju/errors"
 	"github.com/urfave/cli"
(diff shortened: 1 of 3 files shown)


flows/codelingo/review/review/review.go
```diff
@@ -41,12 +33,14 @@ func RequestReview(ctx context.Context, req *flow.ReviewRequest, insecure bool)
...
 
 	util.Logger.Debug("sending request to flow server...")
-	replyc, runErrc, err := c.Run(ctx, &flow.Request{Flow: "review", Payload: payload})
+	replyc, runErrc, cancel, err := flowutil.RunFlow("review", payload, func() proto.Message { return &flow.Reply{} })
 	if err != nil {
 		return nil, nil, errors.Trace(err)
 	}
 	util.Logger.Debug("...request to flow server sent. Received reply channel.")
 
+	_ = cancel
```
waigani: debug code? 

If we're not using cancel, assign the returned value directly to "_" on line 36

sdk/flow/request.go
```diff
@@ -86,3 +94,58 @@ func MarshalChan(replyc chan *grpcflow.Reply, newItem func() proto.Message) (cha
 
 	return itemc, errc
 }
+
+// splitSetters puts user variable setters on their own channel
```
waigani: The term "Setter" is a bit abstract. I'm wondering if we we can use "UserVar" instead? I realise this is late feedback - feel free to ignore it.

BlakeMScurr: Ah yeah, good point. It started as variables.Setter, but I should have used a more explicit term when referring to it in other packages. I'll see if I can fix it quickly.

PR #279, opened 2018-12-14
GitHub comments WIP

BlakeMScurr:
Build suggestion comments. Waiting on dev cluster to test end to end.


flows/codelingo/rewrite/rewrite/cli.go
```diff
@@ -50,6 +50,10 @@ var CLIApp = &flowutil.CLIApp{
 				Name:  "debug",
 				Usage: "Display debug messages",
 			},
+			cli.StringFlag{
+				Name:  "github-comments, g",
```
waigani: Is the json format github specific? If not, it'd be better to call this `dump-comments`. We can then reuse the same dump for other platforms.

PR #280, opened 2018-12-14
Fix client side interactive comment issues.

BlakeMScurr:
Put the value of the output flag on every hunk that's passed to the writer. It's hacky, but much faster than refactoring the flows.

Various other fixes.

### DIFF
--- a/flows/codelingo/rewrite/rewrite/writer.go
+++ b/flows/codelingo/rewrite/rewrite/writer.go
@@ -202,7 +202,7 @@ func newFileSRC(ctx *cli.Context, hunk *rewriterpc.Hunk, fileSRC []byte) ([]byte
 			commentedLine := commentedLines[lineNumber]
 			if updatedLine != commentedLine {
 				c = &comment{
-					Content: fullSuggestion(string(commentedLine)),
+					Content: string(commentedLine),
 					Line:    lineNumber + 1,
 				}
 				break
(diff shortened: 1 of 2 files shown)


flows/codelingo/rewrite/rewrite/writer.go
```diff
@@ -209,6 +212,11 @@ func newFileSRC(ctx *cli.Context, hunk *rewriterpc.Hunk, fileSRC []byte) ([]byte
 	return rewrittenFile, c, nil
 }
 
+// TODO: add review/rewrite comment to template
+func fullSuggestion(newContent string) string {
+	return fmt.Sprintf("```suggestion\n%s\n```\nEdit the suggestion to use an alternative", newContent)
```
waigani: It's not great that ```suggestion is hard coded here - as it's github specific.

sdk/flow/request.go
```diff
@@ -85,6 +83,8 @@ func MarshalChan(replyc chan *grpcflow.Reply, newItem func() proto.Message) (cha
 				continue
 			}
 
+			item = setDefaults(item)
```
waigani: why not `itemc <- setDefaults(item)` ?

flows/codelingo/rewrite/rewrite/writer.go
BlakeMScurr: I can just remove this again and let Shane deal with it when building github comments.

sdk/flow/request.go
BlakeMScurr: Are you suggesting another method like `MarshalChan` that sits downstream? Like `completeItemC<- setDefaults(itemc)`?

BlakeMScurr: Oh, yeah, gotcha, I can just do it in one line haha

Issue #240, opened 2018-11-22
run rewrite unbackslashing results in broken code


mullikine:
https://github.com/mullikine/mullikine.github.io/blob/master/codelingo/troubleshooting/tenets/sprintf_need-multiple-rewrites-and-unbackslash-rewrite.org

PR #285, opened 2018-12-18
Add rewrite to errorf tenet.

BlakeMScurr:


mullikine:
I think this just needs an expected.json file

```
[
  {
   "Comment": "Should replace errors.New(fmt.Sprintf(...)) with errors.Errorf(...).",
   "Filename": "example.go",
   "Line": 16,
   "Snippet": "func doStuffA(param string) error {\n\t// bad\n\treturn errors.New(fmt.Sprintf(\"Don't call with \\\"%s\\\" param - it will literally do nothing!\", param))\n}\n"
  }
 ]
```

mullikine:
LGTM. I'm happy with this workaround. I tested with a string literal ending in a full stop (which was causing problems in the broken tenet).

Please remove the issue https://github.com/codelingo/codelingo/issues/240 as a blocker from issues.yaml and https://github.com/codelingo/codelingo/blob/pr-285/registry/tenets.yaml

PR #296, opened 2019-01-02
Start PSR1 bundle and add tenet to check for camelCase method names

Daanikus:


tenets/codelingo/psr1/camel-case-method-name/expected.json
```diff
@@ -0,0 +1,8 @@
+[
+  {
+   "Comment": "Method names MUST be declared in camelCase",
+   "Filename": "example.php",
+   "Line": 7,
+   "Snippet": "  }\n\n  function HelloOtherWorld() {\n    print \"another hello\";\n  }"
```
BlakeMScurr: Line 7 is still in camel case, it just starts with an upper cased letter. I know PSR isn't that detail though. Here's a comment that goes into a little detail on naming convention https://softwareengineering.stackexchange.com/a/149321. They do suggest methods should be camel case starting with lower case.

tenets/codelingo/psr1/camel-case-method-name/codelingo.yaml
```diff
@@ -0,0 +1,17 @@
...
+        comment: Method names MUST be declared in camelCase
+    query: |
+      import codelingo/ast/php
+
+      php.stmt_class(depth = any):
+        php.stmt_classmethod:
+          @review comment
+          php.identifier:
+            name as methodName
+            regex(/^[A-Z]/, methodName)
```
BlakeMScurr: Shouldn't you be looking for non alphabetic separators instead? Like underscores and dashes etc.

PR #306, opened 2019-01-18
Add new tenet to check for the use of time.After() in switch in loop

Daanikus:


tenets/codelingo/go/use-ticker-in-for-switch/expected.json
```diff
@@ -0,0 +1,8 @@
+[
+  {
+   "Comment": "Make one ticker as opposed to making a new timer each iteration.",
```
waigani: The user has no idea what "ticker" is.

waigani: "Create a ticker outside of the `for` loop with `time.NewTicker(<time period>)` and use it here to avoid creating a new object in memory on each iteration of the loop."

PR #316, opened 2019-02-06
Improve tenet to ignore acronyms and PascalCase

Daanikus:


tenets/codelingo/code-review-comments/go-error-fmt/codelingo.yaml
```diff
@@ -3,9 +3,26 @@ funcs:
...
+
+        var first = a.split(" ")[0]
+        if (first.substring === first.toUpperCase()) {
+          return false
+        }
+
+        // We ignore errors with words containing capitals anywhere in first word
+        // unless the only capital is at the start. This prevents acronyms or
+        // (most) code-related words being caught
+        first = first.substring(1)
```
BlakeMScurr: I'm confused (probably because I don't know JS 😛 ) but is `first.substring` a function, as it appears to be here, or is it a character property, as it appears to be on line 11?

tenets/codelingo/code-review-comments/go-error-fmt/example.go
```diff
@@ -1,8 +1,12 @@
...
 package main
 
+import "fmt"
+
 func main() {
 	fmt.Errorf("something bad.")
 	fmt.Errorf("this is okay")
 	fmt.Errorf("This is not okay")
+	fmt.Errorf("THIS is okay")
+	fmt.Errorf("THIS is not okay.")
```
BlakeMScurr:
```suggestion
	fmt.Errorf("API call thing.")
	fmt.Errorf("THIS is not okay.")
```

wouldn't this currently rewrite to `aPI call thing.`? That's surely bad.

Daanikus: That line is not okay because of the trailing full stop. That line will only be changed to remove that. "THIS" will remain as-is

Daanikus: Is that what you meant?

tenets/codelingo/code-review-comments/go-error-fmt/codelingo.yaml
Daanikus: Whoops. No idea how substring ended up on line 11. It is a string function. Fixed

PR #320, opened 2019-02-11
Fix tenet comments

Daanikus:
An example.go update and a few stylistic tweaks here too

### DIFF
--- a/tenets/codelingo/code-review-comments/avoid-renaming-imports/codelingo.yaml
+++ b/tenets/codelingo/code-review-comments/avoid-renaming-imports/codelingo.yaml
@@ -19,13 +19,13 @@ tenets:
       codelingo/docs:
         title: Avoid Renaming Imports
         body: |
-          Avoid renaming imports except to avoid a name collision; good package names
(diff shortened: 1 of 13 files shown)


tenets/codelingo/code-review-comments/avoid-renaming-imports/codelingo.yaml
```diff
@@ -19,13 +19,14 @@ tenets:
...
-          Avoid renaming imports except to avoid a name collision; good package names
-          should not require renaming. In the event of collision, prefer to rename the
+          Avoid renaming imports except to avoid a name collision. Good package names
+          should not require renaming. In the event of a collision, prefer to rename the
           most local or project-specific import.
       codelingo/review:
         comment: |
-          This import is renamed to `{{importName}}` but this is a bad practice. 
-          Import without specifying a new name.
+          Avoid renaming imports. In the event of a collision, prefer to rename the
```
waigani: Avoid renaming imports, unless there is a collision, in which case prefer to rename the most local or project-specific import.

PR #325, opened 2019-02-13
Exclude tests from comment-first-word-as-subject.

BlakeMScurr:
Users typically don't care about conventional comments on test functions. Nor does godoc, in fact.


rstorr:
@BlakeMScurr prod is back up, should be able to test it.

PR #327, opened 2019-02-14
Fix PSR2 elseif-not-else-if tenet to properly rewrite

Daanikus:


tenets/codelingo/psr2/elseif-not-else-if/codelingo.yaml
```diff
@@ -23,10 +30,7 @@ tenets:
...
 
       @review comment
-      # This does not work as intended.
-      # It removes everything after 'if'
-      # and before closing brace.
-      # Waiting on ability to edit raw file.
-      @rewrite --replace "elseif"
+      @rewrite --replace "{{ extractRaw(rawCode) }}"
       php.stmt_else(depth = any):
+        raw as rawCode
```
BlakeMScurr: Sorry, I don't quite understand. What does this do? It looks like you're replacing all of the spaces inside the else block. Such that this:

```go
else if {
   a := b + c
}
```

becomes:
```
elseif {
   a:=b+c
}
```

Daanikus: @BlakeMScurr One of the quirks of JS. The replace method only replaces the first occurrence. So it only changes else if to elseif.

PR #331, opened 2019-02-18
New tenet update-comment-first-word-as-subject

mullikine:
The new tenet should have fewer false positives.

### DIFF
--- a/registry/tenets.yaml
+++ b/registry/tenets.yaml
@@ -64,6 +64,8 @@ codelingo:
         hasIssues: false
       comment-first-word-as-subject:
         hasIssues: false
+      update-comment-first-word-as-subject:
+        hasIssues: false
       comment-first-word-when-empty:
         hasIssues: false
       good-package-name:
--- a/tenets/codelingo/effective-go/update-comment-first-word-as-subject/codelingo.yaml
+++ b/tenets/codelingo/effective-go/update-comment-first-word-as-subject/codelingo.yaml
@@ -53,7 +53,7 @@ funcs:
           return s.replace(new RegExp("^ *", "i"), "")
         }
 
-        function prepend_name_and_maybe_colonify(s, name) {
+        function maybe_prepend_name(s, name) {
           var carry_on = false;
 
           plurifiable_verbs.forEach(function(verb) {
@@ -71,7 +71,7 @@ funcs:
           if (carry_on) {
               return name + " " + s;
           } else {
-              return name + ": " + s;
+              return s;
           }
         }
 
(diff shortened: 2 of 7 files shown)


registry/tenets.yaml
```diff
@@ -64,6 +64,8 @@ codelingo:
         hasIssues: false
       comment-first-word-as-subject:
         hasIssues: false
+      update-comment-first-word-as-subject:
```
BlakeMScurr: Interesting name - is that just because that's this is the updated version, or because this tenet is intended to do updating/rewriting? I reckon we should have a more descriptive name in either case!

tenets/codelingo/effective-go/update-comment-first-word-as-subject/codelingo.yaml
```diff
@@ -0,0 +1,311 @@
+funcs:
+  - name: isNotATestFile
+    type: asserter
+    body: |
+      function(filename) {
+        // ES5 workaround
+        // TODO: don't use ES5
+        String.prototype.endsWith = String.prototype.endsWith || function(suffix) {
```
Daanikus: Is this some kind of wizardry to detect whether we support ES6 or not?

registry/tenets.yaml
mullikine: To suggest that it updates outdated comments! Hmm I gave it a few minutes thought but this is the best I could think of

tenets/codelingo/effective-go/update-comment-first-word-as-subject/codelingo.yaml
```diff
@@ -0,0 +1,311 @@
...
+        // Replaces the first word with a new function name
+        function updateOld(s, name) {
+          var words = s.split(" ")
+          words[0] = name;
+          return words.join(" ")
+        }
+
+        // modal_verbs etc.
+        var verbs = ["will", "does", "is", "exist"];
+        var plurifiable_verbs = ["split", "purge", "remove", "test", "ensure", "replace", "get", "start",
```
Daanikus: Should add "create" to this list. Ran into that a few times.

tenets/codelingo/effective-go/update-comment-first-word-as-subject/codelingo.yaml
```diff
@@ -0,0 +1,311 @@
...
+        // Do nothing if comment starts with
+        // - non-alphanumerics
+        // - test
+        // - LABEL:
+        // - TODO
+        if ( 
+            (! testre("^[A-Za-z]", commentString)) ||
+            (testre("^test", lc(commentString))) ||
+            (testre("^[a-zA-Z]+:", commentString)) ||
+            (testre("^TODO ", commentString)) ||
```
Daanikus: I don't think this would catch `TODO(Dan)`

registry/tenets.yaml
BlakeMScurr: Great! Naming is worth a couple of minutes, but not worth getting stuck on. It's clear enough.

tenets/codelingo/effective-go/update-comment-first-word-as-subject/codelingo.yaml
mullikine: @BlakeMScurr May have made this comment. ES6 is preferable to ES5 if we want to remove hacky code such as this and replace it with more terse code

PR #348, opened 2019-03-26
Fix bug in tenet

Daanikus:


tenets/codelingo/psr2/elseif-not-else-if/codelingo.yaml
```diff
@@ -3,6 +3,9 @@ funcs:
     type: resolver
     body: |
       function(rawCode) {
+        if (rawCode.substring(0, 7) !== "else if") {
+          return rawCode
```
BlakeMScurr: @Daanikus when would this ever happen? We are only using this function on an else with an if child.

Daanikus: @BlakeMScurr This is a hack as the following two snippets are represented by the same CLQL query:

```
else {
  if (true) {
  }
}
```

```
else if (true) {
}
```

This check just lets us distinguish between the two.

leilaes: @Daanikus @Daanikus if I change `else if` https://github.com/codelingo/codelingo/blob/6177ac37edbec4058faaa5ac09d550e83c578f3a/tenets/codelingo/psr2/elseif-not-else-if/example.php#L8 

to `else` &nbsp;&nbsp;&nbsp;   `if` (more than one space), the tenet doesn't work

PR #349, opened 2019-03-27
Add PSR2 tenet to find and fix non-lowercase language constants

Daanikus:
From PSR2:

## 2.5. Keywords and True/False/Null
PHP keywords MUST be in lower case.

The PHP constants true, false, and null MUST be in lower case.

### DIFF
--- a/tenets/codelingo/psr2/lower-case-constant-values/example-orig.php
+++ b/tenets/codelingo/psr2/lower-case-constant-values/example-orig.php
@@ -1,0 +1,15 @@
+<?php
+
+function helloWorld() {
+    print "hello world!";
+    $thing = true;
+    $thingAgain = true;
+    $thingTheThird = true;
+    $otherThing = false;
+    $another = false;
+    $andAnother = false;
+    $aNull = null;
+    $anotherNull = null;
+    $thirdNull = null;
+    
+}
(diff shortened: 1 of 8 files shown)


tenets/codelingo/psr2/lower-case-constant-values/example-orig.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+function helloWorld() {
+    print "hello world!";
+    $thing = true;
```
mullikine: Should these examples be bad? I get no issues found with the example source

Daanikus: Fixed

tenets/codelingo/psr2/lower-case-constant-values/orig.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+function helloWorld() {
+    print "hello world!";
+    $thing = True;
```
mullikine: Just noticed this. Odd that I got no issues found.

Daanikus: I just changed the file

Daanikus: Oh, nah this file is ignored in .codelingoignore

PR #352, opened 2019-04-01
Add shorthand-assignment-operator tenet

leilaes:
raw code has been used in this tenet as there is no value for scalar_lnumber or scalar_dnumber etc
https://trello.com/c/aLCD4Hto/2365-there-is-no-value-property-for-scalarlnumber-or-scalardnumber-in-php-parser


BlakeMScurr:
@CodeLingoBot review

PR #370, opened 2019-04-11
Add context struct tenet

Daanikus:


tenets/codelingo/code-review-comments/context-in-struct/codelingo.yaml
```diff
@@ -0,0 +1,36 @@
+tenets:
+  - name: context-in-struct
+    flows:
+      codelingo/review:
+        comment: Don't add a Context member to a struct type; instead add a ctx parameter to each method on that type that needs to pass it along.
```
BlakeMScurr: The comment should link to CRC.

BlakeMScurr: We need to justify our standard with an external reference so we don't get pulled into arguments.

tenets/codelingo/code-review-comments/context-in-struct/codelingo.yaml
```diff
@@ -0,0 +1,36 @@
...
+      import codelingo/ast/go
+
+      go.struct_type(depth = any):
+        go.field_list:
+          go.field:
+            go.names
+            @review comment
+            go.selector_expr:
+              go.ident:
+                name == "context"
```
BlakeMScurr:
```suggestion
                # It's unnecessary to confirm that this refers to the actual context package, as any similarly named package should conform to the same requirements
                name == "context"
```

PR #371, opened 2019-04-11
Add full text of Code Review Comments.

BlakeMScurr:
This will be annotated with the tenets that enforce each rule. Every tenet in this bundle should have an annotation.


leilaes:
LGTM

PR #375, opened 2019-04-11
Add named returns tenet

Daanikus:
https://trello.com/c/zUDrp24d


tenets/codelingo/code-review-comments/named-result-params/main.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// GetCount returns the count field of the Thing
+func (e *Example) GetCount() (count int) {
+	count = e.Count
+	return
+}
+
+// GetAll returns all the Thing
+func (e *Example) GetAll() (thing Thing, count int) {
```
BlakeMScurr:
```suggestion
func (e *Example) GetAll() (thing Thing, count int) { // ISSUE
```

tenets/codelingo/code-review-comments/named-result-params/main.go
```diff
@@ -0,0 +1,44 @@
...
+	thing = *e.TheThing
+	count = e.Count
+	return
+}
+
+// Extract a Thing
+func Extract(e *Example) (thing Thing) {
```
BlakeMScurr:
```suggestion
func Extract(e *Example) (thing Thing) { // ISSUE
```

tenets/codelingo/code-review-comments/named-result-params/main.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// GetThing returns the Example's Thing
+func (e *Example) GetThing() (thing *Thing) {
+	thing = e.TheThing
+	return
+}
+
+// GetThingAsValue returns the Example's Thing as value
+func (e *Example) GetThingAsValue() (thing Thing) {
```
BlakeMScurr:
```suggestion
func (e *Example) GetThingAsValue() (thing Thing) { // ISSUE
```

tenets/codelingo/code-review-comments/named-result-params/main.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// Example is an example
+type Example struct {
+	Count    int
+	TheThing *Thing
+}
+
+// GetThing returns the Example's Thing
+func (e *Example) GetThing() (thing *Thing) {
```
BlakeMScurr:
```suggestion
func (e *Example) GetThing() (thing *Thing) { // ISSUE
```

Makes it easier to read.

tenets/codelingo/code-review-comments/named-result-params/codelingo.yaml
```diff
@@ -0,0 +1,68 @@
...
+    type: asserter
+    body: |
+      function(a, b) {
+        return a.toLowerCase() === b.toLowerCase()
+      }
+tenets:
+  - name: named-result-params
+    flows:
+      codelingo/review:
+        comment: Consider what it will look like in godoc. {{returnName}} {{returnType}} will stutter in godoc; better to use {{returnType}}. From [Code Review Comments](https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters)
```
BlakeMScurr:
```suggestion
        comment: Avoid result params with names that are the same as their type. From [Code Review Comments](https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters)
```

I wasn't able to understand how godoc stutters - I didn't manage to run it, so I couldn't verify if your stuttering sentence makes sense. Can you clarify? If you basically copy pasted, then let's just rely on the link.

tenets/codelingo/code-review-comments/named-result-params/codelingo.yaml
```diff
@@ -0,0 +1,68 @@
...
+            go.field:
+              child_count == 2
+              go.names:
+                go.ident:
+                  name as returnName
+              any_of:
+                go.star_expr:
+                  go.ident:
+                    name as returnType
+                    isSameName(returnName, returnType)
```
BlakeMScurr: This seems like a very specific case. Where is it stipulated in CRC that this only applies where the names are the same?

tenets/codelingo/code-review-comments/named-result-params/codelingo.yaml
```diff
@@ -0,0 +1,68 @@
...
+          
+          Naked returns are okay if the function is a handful of lines. Once it's a medium sized function, be explicit with your return values. Corollary: it's not worth it to name result parameters just because it enables you to use naked returns. Clarity of docs is always more important than saving a line or two in your function.
+
+          Finally, in some cases you need to name a result parameter in order to change it in a deferred closure. That is always OK.
+    query: |
+      import codelingo/ast/go
+
+      go.func_decl(depth = any):
+        go.func_type:
+          go.field_list:
```
BlakeMScurr:
```suggestion
          # TODO: add exception for multiple indistinguishable returns values of the same type
          # TODO: add exception for shorter functions
          go.field_list:
```

tenets/codelingo/code-review-comments/FULL_TEXT.md
```diff
@@ -472,6 +472,8 @@ Clarity of docs is always more important than saving a line or two in your funct
 Finally, in some cases you need to name a result parameter in order to change
 it in a deferred closure. That is always OK.
 
+[Tenet](https://github.com/codelingo/codelingo/tree/master/tenets/codelingo/code-review-comments/named-result-params)
```
BlakeMScurr: Is this tenet supposed to address the whole paragraph? It seems limited to a single case where type and name are equal. If it's limited, we should move the annotation somewhere sensible. Otherwise we need to state exceptions or TODOs for exceptions.

tenets/codelingo/code-review-comments/named-result-params/codelingo.yaml
Daanikus: @BlakeMScurr Yeah that was copy/pasted from the doc. I assume it just means it looks silly. This is output of godoc if curious:

```
PACKAGE DOCUMENTATION

package somepackage
    import "somepackage"


TYPES

type Example struct {
    Count    int
    TheThing *Thing
}

    Example is an example

func (e *Example) GetAll() (thing Thing, count int)
    GetAll returns all the Thing

func (e *Example) GetCount() (count int)
    GetCount returns the count field of the Thing

func (e *Example) GetThing() (thing *Thing)
    GetThing returns the Example's Thing

func (e *Example) GetThingAsValue() (thing Thing)
    GetThingAsValue returns the Example's Thing as value

type Thing struct {
    Name   string
    Number int
}

    Thing is an example

func Extract(e *Example) (thing Thing)
    Extract a Thing

```

tenets/codelingo/code-review-comments/named-result-params/codelingo.yaml
Daanikus: Pretty sure that's what it's implying. The stuttering being the same word twice. See doc in yaml file.

tenets/codelingo/code-review-comments/named-result-params/codelingo.yaml
BlakeMScurr: @Daanikus Huh, I assumed it meant that the sentence would be long and unwieldy, but it doesn't look like it even includes the type. I don't think it looks that silly. What you do you find silly about it?

How did you you generate this btw?

tenets/codelingo/code-review-comments/FULL_TEXT.md
Daanikus: Moved

tenets/codelingo/code-review-comments/named-result-params/codelingo.yaml
BlakeMScurr: OK, I figured they just did that because it was easier. It doesn't say that they need to be similar explicitly, does it?

Also, this won't catch `(err error)`.

BlakeMScurr: OK, I figured they just did that because it was easier. It doesn't say that they need to be similar explicitly, does it?

Also, this won't catch `(err error)`.

tenets/codelingo/code-review-comments/named-result-params/codelingo.yaml
Daanikus: @BlakeMScurr I don't necessarily find it silly, I just assume that's what the spec is referring to. I think I've seen it mentioned before about having named returns sharing the name with their type and to avoid it.

Doc was generated by running `godoc somepackage`. It's gotta be in a dir with the same name as package and be in GOPATH.

tenets/codelingo/code-review-comments/named-result-params/codelingo.yaml
Daanikus: No, it won't catch that but that is intentional as err error is common and often used for returning errors from deferred functions. Thought it would be best to exclude that one.

BlakeMScurr: I don't have a strong opinion on whether it should be excluded. But the doc specifically says it would be better to use `error`, and I think we should editorialise as little as possible https://github.com/codelingo/codelingo/pull/375/files#diff-e825bc2f5888babf8dfaf4dae65f36e9R27

PR #377, opened 2019-04-11
Crc package name tenet

Daanikus:


tenets/codelingo/code-review-comments/avoid-meaningless-package-names/codelingo.yaml
```diff
@@ -0,0 +1,39 @@
...
+          "api",
+          "types",
+          "interfaces",
+        ]
+
+        for (var i = 0; i < blacklist.length; i++) {
+          if (name === blacklist[i]) {
+            return true
+          }
+        }
```
BlakeMScurr: Do we have `blacklist.includes(name)` in our JS?

Daanikus: @BlakeMScurr I believe includes() is ES6, which we do not use unfortunately :(

PR #378, opened 2019-04-11
Add tenet to check receiver type of methods for structs containing mutex

Daanikus:


tenets/codelingo/code-review-comments/structs-with-mutex/codelingo.yaml
```diff
@@ -0,0 +1,33 @@
...
+                  name == "sync"
+                go.ident:
+                  name == "Mutex"
+      go.func_decl(depth = any):
+        go.field_list:
+          go.field:
+            go.names:
+              go.ident
+            @review comment
+            go.ident:
```
BlakeMScurr:
```suggestion
            go.ident: // ident is not a child of a star_expr, so it is not a pointer receiver
```
Do we need some kind of exclude here to confirm the receiver is not a pointer? I'm assuming that pointers receivers will have some kind binary expression in between the `go.field` and `go.ident` but this is not obvious. Comment should help.

We could make this even more explicit with `include/exclude` or `path`.

tenets/codelingo/code-review-comments/structs-with-mutex/codelingo.yaml
```diff
@@ -0,0 +1,33 @@
...
+      codelingo/review:
+        comment: If the receiver is a struct that contains a sync.Mutex or similar synchronizing field, the receiver must be a pointer to avoid copying. From [Code Review Comments](https://github.com/golang/go/wiki/CodeReviewComments#receiver-type)
+      codelingo/docs:
+        title: Structs with Mutex
+        body: If the receiver is a struct that contains a sync.Mutex or similar synchronizing field, the receiver must be a pointer to avoid copying.
+    query: |
+      import codelingo/ast/go
+
+      
+      go.type_spec(depth = any):
```
BlakeMScurr: We should scope this to a single package as "You cannot define a method on a type from another package". [cited from some random blog thing](http://www.golangbootcamp.com/book/methods).

tenets/codelingo/code-review-comments/structs-with-mutex/main.go
```diff
@@ -0,0 +1,19 @@
...
+
+import "sync"
+
+type thing struct {
+	name  string
+	num   int
+	mutex sync.Mutex
+}
+
+func (t thing) GetName() string {
```
BlakeMScurr:
```suggestion
func (t thing) GetName() string { // ISSUE
```

tenets/codelingo/code-review-comments/structs-with-mutex/codelingo.yaml
Daanikus: @BlakeMScurr How would we do that? Like this?

```
go.file(depth = any):
  go.ident:
    name as packageName
  ...
go.file(depth = any):
  go.ident:
    name == packageName
```

tenets/codelingo/code-review-comments/structs-with-mutex/codelingo.yaml
Daanikus: @BlakeMScurr Yeah, the ident would have a go.star_expr as a parent. That comment should be enough.

tenets/codelingo/code-review-comments/structs-with-mutex/codelingo.yaml
```diff
@@ -0,0 +1,38 @@
...
+            go.field_list:
+              go.field:
+                go.names:
+                  go.ident
+                go.selector_expr:
+                  go.ident:
+                    name == "sync"
+                  go.ident:
+                    name == "Mutex"
+      go.file(depth = any):
```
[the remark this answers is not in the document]
Daanikus: Ah right, forgot about the dir fact

PR #380, opened 2019-04-12
Add receiver-name-short-reflect-identity tenet

leilaes:


tenets/codelingo/code-review-comments/receiver-name-short-reflect-identity/codelingo.yaml
```diff
@@ -0,0 +1,41 @@
...
+  - name: isNotAbbr
+    type: asserter
+    body: |
+      function(receiverName, typeName) {
+        rc = receiverName.toLowerCase()
+        tn = typeName.toLowerCase()
+        if(rc.length > 2) 
+          return true
+        for (i = 0; i < rc.length; i++) {
+          if(tn.indexOf(rc[i]) == -1)
```
Daanikus: Maybe make a triple equals?

tenets/codelingo/code-review-comments/receiver-name-short-reflect-identity/test.go
```diff
@@ -0,0 +1,35 @@
...
+
+import "fmt"
+
+type TestStruct struct{}
+
+func (t TestStruct) Method1() {
+	fmt.Println("Hello from Method1")
+}
+
+func (ts TestStruct) Method2() {
```
Daanikus: I would've thought that "ts" is acceptable because it is the initials of Test Struct, but it actually passes because the substring "ts" is present in the string "teststruct". What is the thinking behind accepting "ts" as a valid receiver name?

leilaes: I believe that is enough for a two letter string to reflect the identity. The user can choose which two letters which might not necessarily be the first two letters or the initials of a two-word camel case type.

PR #383, opened 2019-04-15
Add camel-case-constants tenet

leilaes:


tenets/codelingo/code-review-comments/camel-case-constants/codelingo.yaml
```diff
@@ -0,0 +1,55 @@
...
+      go.gen_decl(depth = any):
+        tok == "const"
+        go.value_spec:
+          
+          go.names:
+            @review comment
+            @rewrite --replace "{{fixName(varName, varPrivate)}}"
+            go.ident:
+              name as varName
+              hasUnderscore(varName)
```
Daanikus: We would also want to catch things like `const VALUE = 5`. Can you please add a check for that too?

Daanikus: Maybe only check that on words > 4 in length, so we don't catch acronyms like URL

leilaes: `VALUE` is one word, why would we want to change that? I thought the rule is for multi words.

PR #387, opened 2019-04-17
Add no-get-in-getters-name tenet

leilaes:


tenets/codelingo/effective-go/no-get-in-getters-name/codelingo.yaml
```diff
@@ -0,0 +1,42 @@
...
+    body: |
+      function(name) {
+        return name.substring(0,3) === "Get"
+      }
+
+  - name: fixName
+    type: resolver
+    body: |
+      function(name) {
+        return name.charAt(0).toUpperCase() + name.substring(1)
```
Daanikus:
```suggestion
        name = name.substring(3)
        return name.charAt(0).toUpperCase() + name.substring(1)
```

Daanikus: Does that look right? We wanna strip the "Get" I assume. And missing rewrite decorator. Have you tested the rewrite or are you postponing it?

Daanikus: Callgraph might work for this

leilaes: `name` here is the "field" of the struct, not Getter method name and wouldn't have "Get" in it. I have not tried rewrite yet and have postponed it for now. Is it priority?

Daanikus: No, not a priority. You got a LGTM so you can leave it

PR #388, opened 2019-04-17
Add initialize-using-composite-literal tenet

leilaes:


tenets/codelingo/effective-go/loop-variable-used-in-go-routine/test.go
```diff
@@ -0,0 +1,79 @@
...
+			if n {
+				fmt.Println(val)
+			}
+		}(val, n)
+	}
+}
+
+func Test6(values []int, n bool) {
+	for val := range values {
+		go func(n bool) { //ISSUE
```
Daanikus: What's wrong with this one?

tenets/codelingo/effective-go/loop-variable-used-in-go-routine/codelingo.yaml
```diff
@@ -0,0 +1,95 @@
+funcs:
+  - name: isEqualtoAny
+    type: asserter
+    body: |
+      function(name1, name2, name3) {
+        return (name1 == name3 && name1 != "_" ) || (name2 == name3 && name2 != "_" ) }
```
Daanikus: Move `}` to new line

tenets/codelingo/effective-go/loop-variable-used-in-go-routine/codelingo.yaml
```diff
@@ -0,0 +1,95 @@
...
+
+  - name: isEqual
+    type: asserter
+    body: |
+      function(name1, name2) {
+        return name1 == name2 && name1 != "_"  }
```
Daanikus: Move `}` to new line

tenets/codelingo/effective-go/loop-variable-used-in-go-routine/test.go
leilaes: Because `val` is used in the goroutine and not passed.

Daanikus: Ah right. I thought the comment was saying the issue was with the bool.

PR #395, opened 2019-04-24
 Add else-if-when-mutually-exclusive tenet

CodeLingoTeam:
Inspired by this PR comment https://github.com/mssola/user_agent/pull/41/files/edb5c17fc5816d71144ecaa5ea7c2d4ba1cecd52#r153998505


tenets/codelingo/mssola-user_agent/else-if-when-mutually-exclusive/expected.json
```diff
@@ -0,0 +1,8 @@
+[
+  {
+   "Comment": "This is a function, but you probably already knew that.",
```
Daanikus: Maybe regen with updated comment?

PR #397, opened 2019-04-25
Retry on incorrect option

Daanikus:
It was annoying that the program would exit on a typo, especially during a large review process.


sdk/flow/confirm.go
```diff
@@ -51,7 +52,8 @@ func (item *ConfirmerItem) Confirm(ctx *cli.Context) (bool, error) {
 	fmt.Scanln(&option)
 	action, err := item.action(option)
 	if err != nil {
-		return false, errors.Trace(err)
+		fmt.Fprintf(os.Stderr, "no action found for option %q", option)
```
BlakeMScurr:
```suggestion
		util.Logger.Warnf("no action found for option %q", option)
```

From `"github.com/codelingo/lingo/app/util"`

BlakeMScurr: We should check the error string to make sure it's due to a typo.

Daanikus: I'm pretty sure this is the only possible error returned from that func.

Daanikus: Will this notify the user though? Or just if debug flag is set?

BlakeMScurr: Ah OK, yeah I missed the point here.

PR #402, opened 2019-04-30
Add doc flow to tenet

Daanikus:


tenets/codelingo/code-review-comments/declare-empty-slice/codelingo.yaml
```diff
@@ -12,9 +12,29 @@ tenets:
...
+          t := []string{}
+          ```
+
+          The former declares a nil slice value, while the latter is non-nil but zero-length. They are functionally equivalent—their `len` and `cap` are both zero—but the nil slice is the preferred style.
+
+          Note that there are limited circumstances where a non-nil but zero-length slice is preferred, such as when encoding JSON objects (a `nil` slice encodes to `null`, while `[]string{}` encodes to the JSON array `[]`).
+
+          When designing interfaces, avoid making a distinction between a nil slice and a non-nil, zero-length slice, as this can lead to subtle programming errors.
+
+          For more discussion about nil in Go see Francesc Campoy's talk [Understanding Nil](https://www.youtube.com/watch?v=ynoY2xz-F8s).
```
BlakeMScurr: Discussion section may be too long for a slide. @Daanikus can you confer with @waigani about that.

PR #409, opened 2019-05-02
Run script to tidy up all lingo_bundle files

Daanikus:
These files are needed to correctly import tenets. Used a script to ensure consistency. Find it at github.com/daanikus/bundlegen


leilaes:
LGTM

PR #413, opened 2019-05-15
Update declare-empty-slice tenet

leilaes:


tenets/codelingo/code-review-comments/declare-empty-slice/codelingo.yaml
```diff
@@ -22,13 +24,13 @@ tenets:
...
 
-          over 
+          over
 
           ```go
           t := []string{}
           ```
 
-          The former declares a nil slice value, while the latter is non-nil but zero-length. They are functionally equivalent—their `len` and `cap` are both zero—but the nil slice is the preferred style.
+          The former declares a nil slice value, while the latter is non-nil but zero-length. They are functionally equivalent ^`^ttheir `len` and `cap` are both zero ^`^tbut the nil slice is the preferred style.
```
BlakeMScurr: What do the carats (`^`) do here?

tenets/codelingo/code-review-comments/declare-empty-slice/codelingo.yaml
```diff
@@ -51,4 +53,4 @@ tenets:
           end_offset
           go.composite_lit:
             child_count == 1
-            go.array_type
+            go.array_type
```
BlakeMScurr: Probably better to have a newline at the end of file - but it's not a big deal either way.

tenets/codelingo/code-review-comments/declare-empty-slice/codelingo.yaml
```diff
@@ -4,7 +4,9 @@ funcs:
     body: |
       function(sliceType) {
         sliceType = sliceType.split(' ').join('');
-        return  sliceType.substring(0, sliceType.indexOf("{"))
+        if (sliceType.substring(sliceType.length - 4, sliceType.length) == "{}{}" && sliceType.indexOf("interface") !== 0)
```
BlakeMScurr: Can you please add a comment explaining this exception? I couldn't quite understand it. Like, why `sliceType.length - 4`, for example.

BlakeMScurr: If it has to do with the fact the `{}` in `interface{}` is part of the type, the same thing applies for `struct{}`.

leilaes: I am checking for empty interface, can you please give me an example of struct and I will include that too

tenets/codelingo/code-review-comments/declare-empty-slice/codelingo.yaml
```diff
@@ -4,7 +4,9 @@ funcs:
     body: |
       function(sliceType) {
         sliceType = sliceType.split(' ').join('');
-        return  sliceType.substring(0, sliceType.indexOf("{"))
+        if (sliceType.substring(sliceType.length - 4, sliceType.length) == "{}{}" && (sliceType.indexOf("interface") !== 0 || sliceType.indexOf("struct") !== 0)) // This check is for empty interface or struct
```
BlakeMScurr:
```suggestion
        // This check is for empty interface or struct, which include `{}` as part of their type definition
        if (sliceType.substring(sliceType.length - 4, sliceType.length) == "{}{}" && (sliceType.indexOf("interface") !== 0 || sliceType.indexOf("struct") !== 0))
```

Readability

PR #414, opened 2019-05-21
Find sync/mutex that can be replaced with sync/atomic

leilaes:


tenets/codelingo/thrasher-gocryptotrader/codelingo.yaml
```diff
@@ -0,0 +1,72 @@
+funcs:
+  - name: isGreater
```
CodeLingoTeam:
```suggestion
  - name: isBPlus2
```

PR #423, opened 2019-06-13
Fix return-discarded-errors tenet's review

BlakeMScurr:
Remove undefined variable.
Move decorator to error assignment.


leilaes:
LGTM

PR #427, opened 2019-07-10
Add path to ReviewRequest cli

leilaes:


flows/codelingo/review/review/cli.go
```diff
@@ -166,6 +167,13 @@ func reviewAction(cliCtx *cli.Context) (chan proto.Message, <-chan *flowutil.Use
 		return nil, nil, nil, nil, errors.Trace(err)
 
 	}
+	wd, err := os.Getwd()
+	if err != nil {
+		return nil, nil, nil, nil, errors.Trace(err)
+
+	}
+	parts := strings.Split(wd, "/")
+	path := parts[2] + "/" + parts[len(parts)-1]
```
BlakeMScurr: I don't think this will work on a machine where `GOPATH` is `/home/users/me/go`, for example; you're assuming `GOPATH` is a certain length. You should explicitly get the go path and trim it from the working directory. If there is no `GOPATH` set then `path` should be `""`. This will make the purpose of the code much clearer too.

BlakeMScurr: Sorry, I misunderstood what you were trying to do. Trying to get the ubuntu username won't work as per our conversation:
```suggestion
	path := parts[len(parts)-2] + "/" + parts[len(parts)-1]
```

BlakeMScurr: However, reviews don't have to run from the root of a repo. So you'll actually have to use the git root rather than the working directory.

The fact that `path` should still be empty when you don't have a `GOPATH` variable still stands. Path only applies to go.

flows/codelingo/review/review/cli.go
```diff
@@ -166,16 +169,67 @@ func reviewAction(cliCtx *cli.Context) (chan proto.Message, <-chan *flowutil.Use
...
-		Repo:     name,
-		Sha:      sha,
-		Patches:  patches,
-		Vcs:      vcsTypeStr,
-		Dir:      workingDir,
-		Dotlingo: dotlingo,
+
+	var req *flow.ReviewRequest
+	// If path is not empty, it is sent in ReviewRequest and used in go lexicon
+	if path != "" {
```
BlakeMScurr: Just return the version with path. Unassigned fields default to the zero value in Go. https://play.golang.org/p/p16l5ISRXBw.

```go
package main

import (
	"fmt"
)

type s struct {
	path string
}

func main() {
	implicit := &s{}
	explicit := &s{
		path: "",
	}
	fmt.Println(implicit.path == explicit.path) // true
}
```

flows/codelingo/review/review/cli.go
```diff
@@ -166,16 +169,67 @@ func reviewAction(cliCtx *cli.Context) (chan proto.Message, <-chan *flowutil.Use
 		return nil, nil, nil, nil, errors.Trace(err)
 
 	}
+	// Build `path` for go projects as "reponame" for a local repo or "github.com/username/reponame" for a github repo
+	// path variable is initially empty and will remain empty for none go repos
```
BlakeMScurr:
```suggestion
	// It is empty for non-Go repos
```

flows/codelingo/review/review/cli.go
```diff
@@ -166,16 +169,67 @@ func reviewAction(cliCtx *cli.Context) (chan proto.Message, <-chan *flowutil.Use
 		return nil, nil, nil, nil, errors.Trace(err)
 
 	}
+	// Build `path` for go projects as "reponame" for a local repo or "github.com/username/reponame" for a github repo
```
BlakeMScurr:
```suggestion
	// Path is the relative path from $GOPATH/src to the repo root. It's required for package resolution.
```

There's no need to mention the value of path for a github repo here, since this is the CLI and it only concerns local repos.

flows/codelingo/review/review/cli.go
```diff
@@ -166,16 +169,67 @@ func reviewAction(cliCtx *cli.Context) (chan proto.Message, <-chan *flowutil.Use
...
+	// path variable is initially empty and will remain empty for none go repos
+	path := ""
+
+	// Find the current working directory, i.e. where user executed `lingo run review` command on the repo
+	wd, err := os.Getwd()
```
BlakeMScurr: You should use the root of the git repo instead, since the review can be run from a subdirectory.

flows/codelingo/review/review/cli.go
```diff
@@ -222,3 +276,23 @@ func reviewAction(cliCtx *cli.Context) (chan proto.Message, <-chan *flowutil.Use
...
 	return resultc, userVarc, errc, cancel, errors.Trace(err)
 }
+
+func isGoProject(currentDir string) (bool, error) {
+	isGo := false
+	items, err := ioutil.ReadDir(currentDir)
+	if err != nil {
+		return false, err
+	}
+	for _, item := range items {
```
BlakeMScurr:
```suggestion
	for _, item := range items {
	if !item.IsDir() && filepath.Ext(item.Name()) == ".go" {
	    return true, nil
	}
```
If you use the recursive method you should exit before recursion, otherwise we'll have to check every single file.

flows/codelingo/review/review/cli.go
```diff
@@ -222,3 +276,23 @@ func reviewAction(cliCtx *cli.Context) (chan proto.Message, <-chan *flowutil.Use
 	resultc, userVarc, errc, err := RequestReview(ctx, req, insecure)
 	return resultc, userVarc, errc, cancel, errors.Trace(err)
 }
+
+func isGoProject(currentDir string) (bool, error) {
+	isGo := false
+	items, err := ioutil.ReadDir(currentDir)
+	if err != nil {
+		return false, err
```
BlakeMScurr:
```suggestion
		return false, errors.Trace(err)
```

flows/codelingo/review/review/cli.go
```diff
@@ -166,16 +169,67 @@ func reviewAction(cliCtx *cli.Context) (chan proto.Message, <-chan *flowutil.Use
...
+	goProject, err := isGoProject(wd)
+	if err != nil {
+		return nil, nil, nil, nil, errors.Trace(err)
+	}
+
+	// Check if GOPATH is set
+	goPath := os.Getenv("GOPATH")
+
+	if goPath != "" && goProject {
+		if strings.Contains(wd, "github.com") { // Check if the working directory includes "github.com/username/reponame"
```
BlakeMScurr: Can you explain the purpose of lines 192-204? I figured you could do something as simple as:

```go
path = strings.TrimPrefix(wd, goPath)
```

flows/codelingo/review/review/cli.go
BlakeMScurr: Please put path generation into its own function; the `reviewAction` function is about 200 lines long!

flows/codelingo/review/review/cli.go
```diff
@@ -222,3 +237,56 @@ func reviewAction(cliCtx *cli.Context) (chan proto.Message, <-chan *flowutil.Use
...
+
+		} else if item.IsDir() && !strings.HasPrefix(item.Name(), ".") {
+			isGo, err = isGoProject(filepath.Join(currentDir, item.Name()))
+			if err != nil {
+				return false, errors.Trace(err)
+			}
+
+		}
+	}
+	return isGo, nil
```
BlakeMScurr:
```suggestion
	return false, nil
```

flows/codelingo/review/review/cli.go
```diff
@@ -222,3 +237,56 @@ func reviewAction(cliCtx *cli.Context) (chan proto.Message, <-chan *flowutil.Use
...
+
+		} else if strings.HasPrefix(repoRoot, goPath+"/src/") {
+			path = strings.TrimPrefix(repoRoot, goPath+"/src/")
+		}
+	}
+	return path, nil
+}
+
+func isGoProject(currentDir string) (bool, error) {
+	isGo := false
```
BlakeMScurr:
```suggestion
```

flows/codelingo/review/review/cli.go
```diff
@@ -222,3 +237,56 @@ func reviewAction(cliCtx *cli.Context) (chan proto.Message, <-chan *flowutil.Use
...
+	items, err := ioutil.ReadDir(currentDir)
+	if err != nil {
+		return false, errors.Trace(err)
+	}
+	for _, item := range items {
+		if !item.IsDir() && filepath.Ext(item.Name()) == ".go" {
+			return true, nil
+
+		} else if item.IsDir() && !strings.HasPrefix(item.Name(), ".") {
+			isGo, err = isGoProject(filepath.Join(currentDir, item.Name()))
```
BlakeMScurr:
```suggestion
			isGo, err := isGoProject(filepath.Join(currentDir, item.Name()))
```

flows/codelingo/review/review/cli.go
```diff
@@ -222,3 +237,56 @@ func reviewAction(cliCtx *cli.Context) (chan proto.Message, <-chan *flowutil.Use
...
+			if err != nil {
+				return false, errors.Trace(err)
+			}
+
```
BlakeMScurr:
```suggestion

			if isGo {
				return true
			}
```

BlakeMScurr: This way we return as soon as we see a go file, rather than continuing to search the rest of its siblings.

flows/codelingo/review/review/cli.go
```diff
@@ -165,17 +169,28 @@ func reviewAction(cliCtx *cli.Context) (chan proto.Message, <-chan *flowutil.Use
 	if err != nil {
 		return nil, nil, nil, nil, errors.Trace(err)
 
+	}
+	// `path` is the relative path from $GOPATH/src to the repo root. It's required for package resolution
```
BlakeMScurr: Make this a function comment.

flows/codelingo/review/review/cli.go
```diff
@@ -165,17 +169,28 @@ func reviewAction(cliCtx *cli.Context) (chan proto.Message, <-chan *flowutil.Use
...
+	// It is empty for non-Go repos
+	path, err := findPath()
+	if err != nil {
+		return nil, nil, nil, nil, errors.Trace(err)
+
 	}
 
 	ctx, cancel := util.UserCancelContext(context.Background())
+
+	// If path is not empty, it is sent in ReviewRequest and used in go lexicon
```
BlakeMScurr:
```suggestion
```

Unnecessary detail.

flows/codelingo/review/review/cli.go
```diff
@@ -222,3 +237,56 @@ func reviewAction(cliCtx *cli.Context) (chan proto.Message, <-chan *flowutil.Use
...
+	if err != nil {
+		return "", errors.Trace(err)
+	}
+
+	// Check if GOPATH is set
+	goPath := os.Getenv("GOPATH")
+
+	// Check if repo is in GOPATH
+	if goPath != "" && goProject {
+		if strings.HasPrefix(repoRoot, goPath+"/src/github.com") {
```
BlakeMScurr: Not necessary. You just need `if strings.HasPrefix(repoRoot, goPath+"/src/")`

flows/codelingo/review/review/cli.go
```diff
@@ -222,3 +235,55 @@ func reviewAction(cliCtx *cli.Context) (chan proto.Message, <-chan *flowutil.Use
 	resultc, userVarc, errc, err := RequestReview(ctx, req, insecure)
 	return resultc, userVarc, errc, cancel, errors.Trace(err)
 }
+
+// `path` is the relative path from $GOPATH/src to the repo root. It's required for package resolution
+// It is empty for non-Go repos
+func findPath() (string, error) {
+	path := ""
```
BlakeMScurr:
```suggestion
```

flows/codelingo/review/review/cli.go
```diff
@@ -222,3 +235,55 @@ func reviewAction(cliCtx *cli.Context) (chan proto.Message, <-chan *flowutil.Use
...
+
```
BlakeMScurr:
```suggestion
```

flows/codelingo/review/review/cli.go
```diff
@@ -222,3 +235,55 @@ func reviewAction(cliCtx *cli.Context) (chan proto.Message, <-chan *flowutil.Use
...
+		return "", errors.Trace(err)
+	}
+
+	// Check if GOPATH is set
+	goPath := os.Getenv("GOPATH")
+
+	// Check if repo is in GOPATH
+	if goPath != "" && goProject {
+		if strings.HasPrefix(repoRoot, goPath+"/src/") {
+			path = strings.TrimPrefix(repoRoot, goPath+"/src/")
```
BlakeMScurr:
```suggestion
			return strings.TrimPrefix(repoRoot, goPath+"/src/")
```

flows/codelingo/review/review/cli.go
```diff
@@ -222,3 +235,55 @@ func reviewAction(cliCtx *cli.Context) (chan proto.Message, <-chan *flowutil.Use
...
+		}
+	}
+	return path, nil
```
BlakeMScurr:
```suggestion
	return "", nil
```

Issue #150, opened 2018-10-25
32bit builds failing


emersonwood:
All 32bit builds are failing and will not be useable by the end user.

```
# github.com/codelingo/codelingo/vendor/github.com/codelingo/clql/inner
../../../vendor/github.com/codelingo/clql/inner/inner_parser.go:1980: constant 4294967294 overflows int
```

BlakeMScurr:
Details:

**Describe the bug**
An integer overflow error in the codelingo/clql dependency cause 32bit flow builds to fail.

**To Reproduce**
Steps to reproduce the behavior:
1. Go to [codelingo/sdk/flow](https://github.com/codelingo/codelingo/tree/master/sdk/flow) in the CLI
2. Run `./buildall.sh`
3. Scroll down to '....'
4. See error

**Expected behavior**
All binaries in the flow folder to update.

**Actual Behavior**

Only 64bit builds succeeded. The command showed build errors:

```
./buildall.sh 
+ path=/home/matt/go
+ [[ -z /home/matt/go ]]
+ codelingoPath=/home/matt/go/src/github.com/codelingo/codelingo
+ v='
windows,386;linux,386;"'
+ version=0.0.0
+ for owner in '$codelingoPath/flows/*/'
+ for d in '$owner*/'
+ cd /home/matt/go/src/github.com/codelingo/codelingo/flows/codelingo/docs/
+ echo 'windows,386;linux,386;"'
+ IFS=,
+ read '-d;' os arch
+ cd /home/matt/go/src/github.com/codelingo/codelingo/flows/codelingo/docs/
+ '[' windows == windows ']'
+ ext=.exe
+ binpath=./bin/windows/386/0.0.0
+ mkdir -p ./bin/windows/386/0.0.0
+ GOOS=windows
+ GOARCH=386
+ go build -o ./bin/windows/386/0.0.0/cmd.exe
# github.com/codelingo/codelingo/vendor/github.com/codelingo/clql/inner
../../../vendor/github.com/codelingo/clql/inner/inner_parser.go:1980: constant 4294967294 overflows int
../../../vendor/github.com/codelingo/clql/inner/inner_parser.go:1980: constant 4160749567 overflows int
../../../vendor/github.com/codelingo/clql/inner/inner_parser.go:1980: constant 4294967295 overflows int
+ cd ./bin/windows/386/0.0.0
+ filename=cmd.exe
+ '[' windows == windows ']'
+ fn=cmd.exe.zip
[...]

PR #433, opened 2019-08-13
Rebuild actions to support flows and actions in yaml

Daanikus:


Daanikus:
Reverted Gopkg changes. Something real funky was going on.

Deleted all search binaries as they are no longer used (and are now just an empty main func).

Removed 32bit binaries due to https://github.com/codelingo/codelingo/issues/150

PR #436, opened 2019-08-21
rebuild docs and remove x86 from build.sh

Daanikus:
Updated CLQL for docs actions. Will fix errors and support both flows and actions.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -86,14 +86,14 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:210ca4edc7c277d3a1becbf65b30e926ef4642b02f96146cc64c0c9f8e80122a"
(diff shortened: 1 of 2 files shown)


Gopkg.lock
```diff
@@ -888,6 +904,7 @@
     "github.com/codelingo/lingo/service",
     "github.com/codelingo/lingo/service/grpc",
     "github.com/codelingo/lingo/vcs",
+    "github.com/codelingo/rectangle",
```
rstorr: Can you look into why we're importing github.com/codelingo/rectangle ? cheers

Daanikus: @rstorr Yeah, it's part of Leila's typed nil tenet. It's used as an example.

rstorr:
Worth checking that our docs don't suggest we support 32bit

PR #439, opened 2019-09-02
Seccomp

BlakeMScurr:
Finds repos that contain Kubernetes configuration files but don't have a policy that requires that pods run in seccomp mode.

Relies on https://github.com/codelingo/platform/pull/1385


tenets/codelingo/kubernetes-best-practices/seccomp/codelingo.yaml
```diff
@@ -0,0 +1,49 @@
+tenets:
+  - name: find-funcs
+    flows:
+      codelingo/review:
+        comment: | 
+          This kubernetes config file is not paired with a PodSecurityPolicy requiring its pods to run in seccomp.
+
+          Pods should be run in `docker/default` or `runtime/default`, but they `unconfined` by default to ensure backwards compatability.
```
Daanikus:
```suggestion
          Pods should be run in `docker/default` or `runtime/default`, but they are `unconfined` by default to ensure backwards compatability.
```

Daanikus:
Also, bundle file and registry edit?

PR #443, opened 2019-09-16
Tenet for breaking a select in a for loop.

BlakeMScurr:
`example.go` is heavily commented and should explain the issue in detail.

### DIFF
--- a/tenets/codelingo/go/break-select-in-for/example.go
+++ b/tenets/codelingo/go/break-select-in-for/example.go
@@ -1,0 +1,102 @@
+package main
+
+import "fmt"
+
+func main() {
+	// We generally write break statements to break out of repetitious things like for loops
+	fmt.Println("Expecting 0, 1:")
+	for i := range []int{0, 1, 2} {
+		fmt.Println(i)
+		if i == 1 {
+			break // Non-issue
+		}
+	}
+
+	// and don't often break from selects, so it's easy to forget that it's possible
+	fmt.Println("Expecting 0:")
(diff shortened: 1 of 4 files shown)


tenets/codelingo/go/break-select-in-for/example.go
```diff
@@ -0,0 +1,102 @@
...
+		fmt.Println(i)
+		if i == 1 {
+			break // Non-issue
+		}
+	}
+
+	// and don't often break from selects, so it's easy to forget that it's possible
+	fmt.Println("Expecting 0:")
+	c1 := make(chan int)
+	go func() {
```
fwereade: FWIW, this is one of the cases where buffered channels make sense and (IMO) reduce confusion by skipping the goroutine: https://play.golang.org/p/uvyrMmaOmFF

fwereade: or indeed https://play.golang.org/p/yg-B1s3KHuN

tenets/codelingo/go/break-select-in-for/example.go
```diff
@@ -0,0 +1,90 @@
...
+	for i := range []int{0, 1, 2} {
+		for j := range []int{0, 1, 2} {
+			fmt.Println(i, j)
+			if j == 0 {
+				break l // Non-issue
+			}
+		}
+	}
+
+	// Which works perfectly fine to break a select insisde a for loop
```
fwereade: s/insisde/inside/

PR #445, opened 2019-09-17
Add nudge for deprecated functions.

BlakeMScurr:


registry/tenets.yaml
```diff
@@ -119,6 +119,8 @@ codelingo:
         hasIssues: false
       break-select-in-for:
         hasIssues: false
+      deprecated:
+        hasIssues: false
```
Daanikus: :tada:

tenets/codelingo/go/deprecated/codelingo.yaml
```diff
@@ -0,0 +1,18 @@
+tenets:
+  - name: deprecated
+    flows:
```
Daanikus: s/flows/actions

Daanikus: Does lingo init still generate with "flows"?

BlakeMScurr: Ooh, yeah it does. And it seems to work too 🤔

Daanikus: We support both, so that's good!

PR #447, opened 2019-09-18
Add ignore goroutine return values tenet

CallanTaylor:


tenets/codelingo/go/ignore-goroutine-return-values/codelingo.yaml
```diff
@@ -0,0 +1,19 @@
...
+      go.go_stmt(depth = any):
+        go.call_expr:
+          any_of:
+            go.func_lit:
+              go.func_type:
+                go.field_list
+                go.field_list
+            edge("calls"):
+              go.func_decl:
+                go.return_stmt(depth = any)
```
BlakeMScurr: I'm guessing this means that if you have a `return` in a `func_decl` we assume that it returns a value? You can return without returning a value. [pg link for](https://play.golang.org/p/gf5aXD7zjY5):

```go
package main

import (
	"fmt"
)

func main() {
	fmt.Println("Hello, playground")
	if true {
		return	
	}
	fmt.Println("Hello, playground")
}
```

Probably safer to use that same pattern as above:

```yaml
go.func_decl:
  go.func_type:
    go.field_list
    go.field_list
```

tenets/codelingo/go/ignore-goroutine-return-values/example.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+func main() {
+	go A() // Issue
+        go B() // Non Issue
+	<-time.After(time.Second)
+
+	go func() string {
+	    word := "Hello"
+	    return word // Issue
```
BlakeMScurr:
```suggestion
	    return word
```

tenets/codelingo/go/ignore-goroutine-return-values/example.go
```diff
@@ -0,0 +1,33 @@
...
+	"time"
+	"fmt"
+)
+
+func main() {
+	go A() // Issue
+        go B() // Non Issue
+	<-time.After(time.Second)
+
+	go func() string {
```
BlakeMScurr:
```suggestion
	go func() string { // Issue
```

tenets/codelingo/go/ignore-goroutine-return-values/example.go
```diff
@@ -0,0 +1,33 @@
...
+	    word := "Hello"
+	    return word // Issue
+        }()
+
+
+
+
+        go func() {
+            fmt.Println("Hello") // Non Issue
```
BlakeMScurr:
```suggestion
            fmt.Println("Hello")
```

tenets/codelingo/go/ignore-goroutine-return-values/example.go
```diff
@@ -0,0 +1,33 @@
...
+
+	go func() string {
+	    word := "Hello"
+	    return word // Issue
+        }()
+
+
+
+
+        go func() {
```
BlakeMScurr:
```suggestion
        go func() { // Issue
```

tenets/codelingo/go/ignore-goroutine-return-values/example.go
BlakeMScurr: Issue comments should go on the lines that'll actually be returned.

tenets/codelingo/go/ignore-goroutine-return-values/example.go
CallanTaylor: I think you meant `Non Issue` here to replace the one you suggested removing in the above comment

PR #448, opened 2019-09-19
Remove dev tenet

Daanikus:
The tenet introduced a dependency for type resolution, so had to run dep


Gopkg.lock
```diff
@@ -116,11 +239,32 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:78d0cea6d598e3f2498ee014cd4ee767966698a35bef29b4dd28922846c5eabb"
-  name = "github.com/codelingo/rectangle"
-  packages = ["."]
+  digest = "1:cec4e8bf3baf27788873f0e0bbaeaec9c806c15ad59f1f2173164a15395cca16"
+  name = "github.com/codelingo/platform"
```
BlakeMScurr: Any idea why we're suddenly dependant on the platform? It doesn't make a lot of sense to me that we have more dependencies since you only removed Go code.

Daanikus: Hmmm.... no I have no idea where this came from...

Daanikus: Ah, looks like someone imported platform in an example file

PR #450, opened 2019-09-20
Ignore defered returns

CallanTaylor:


tenets/codelingo/go/ignore-defered-returns/codelingo.yaml
```diff
@@ -0,0 +1,16 @@
...
+      codelingo/review:
+        comment: Values returned from defered functions should not be ignored.
+    query: |
+      import codelingo/ast/go
+
+      @review comment
+      go.defer_stmt(depth = any):
+        go.call_expr:
+          edge("calls"):
+            go.func_decl:
```
BlakeMScurr: You can defer a `func_lit` too https://play.golang.org/p/dO4HpjzLFBM

tenets/codelingo/go/ignore-defered-returns/codelingo.yaml
```diff
@@ -0,0 +1,22 @@
...
+      @review comment
+      go.defer_stmt(depth = any):
+        go.call_expr:
+          any_of:
+            edge("calls"):
+              go.func_decl:
+                go.func_type:
+                  go.field_list
+                  go.field_list
+            edge("calls"):
```
BlakeMScurr: Does it work with this?

```
edge("calls"):
  any_of:
    go.func_decl
    go.func_lit
```

Rather than:

```
any_of:
  edge("calls"):
    go.func_decl
  edge("calls"):
    go.func_lit
```

Just to save a single line.

BlakeMScurr: In fact, can we even drop the `edge("calls")` for the `func_lit` like in the goroutine tenet?

```
any_of:
  go.func_decl
  edge("calls"):
    go.func_lit
```

I'd argue that:

```
edge("calls"):
  any_of:
    go.func_decl
    go.func_lit
```

is the easiest to understand personally.

CallanTaylor: We cant put `any_of` as a child of `edge("calls")` it seems but you are right about being able to drop the `edge("calls")` for the `func_lit` altogether so Ill do that and merge!

PR #451, opened 2019-09-24
Add tenet and update registry

CallanTaylor:
This Tenet/Standard/Spec catches situations where a function is accessing a struct type with exported fields by reference. This can lead to multiple functions, possibly operating on different go-routines, to make read and writes to the same memory at the same time and is generally an anti-pattern.

### DIFF
--- a/tenets/codelingo/go/pass-struct-by-pointer/codelingo.yaml
+++ b/tenets/codelingo/go/pass-struct-by-pointer/codelingo.yaml
@@ -1,0 +1,26 @@
+tenets:
+  - name: pass-struct-by-pointer
+    actions:
+      codelingo/review:
+        comment: Passing a pointer to a struct with exported fields such as {{ structName }} is bad practice as it can result in multiple process reading and writing to the same memory.
+    query: |
+      import codelingo/ast/go
+
+      go.type_spec(depth = any):
+        go.ident:
+          name as structName
+        go.struct_type:
+          go.field_list:
+            go.field:
+              go.names:
+                go.ident:
+                  name as fieldName
+                  regex(/^[A-Z]/, fieldName)
+      @review comment
+      go.func_decl(depth = any):
+        go.field_list(depth = any):
+          go.field:
+            go.names
+            go.star_expr:
+              go.ident: 
+                name == structName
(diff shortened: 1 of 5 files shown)


tenets/codelingo/go/pass-struct-by-pointer/codelingo.yaml
```diff
@@ -0,0 +1,26 @@
...
+      go.type_spec(depth = any):
+        go.ident:
+          name as structName
+        go.struct_type:
+          go.field_list:
+            go.field:
+              go.names:
+                go.ident:
+                  name as fieldName
+                  regex(/^[A-Z]/, fieldName)
```
Daanikus: Can we not use the public property here? Assuming this is to only include exported fields

tenets/codelingo/go/pass-struct-by-pointer/example.go
```diff
@@ -0,0 +1,44 @@
...
+	age int
+}
+
+type dog struct {
+	name string
+	age int
+}
+
+func getCatName(c *Cat) {
+	fmt.Println(c.Name) // Issue, accessing struct with exported field(s) by reference
```
fwereade: hmm, I think this needs to be more sophisticated

I think the big issue is in *keeping* a reference, such that more than one piece of code could operate on it at the same time

fwereade: to elaborate: in these cases, the calling context can be safely assumed not to operate on it while the function is running, because it's not doing anything until the function returns

tenets/codelingo/go/pass-struct-by-pointer/example.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func getCatAge(c Cat) { // Non Issue, passing struct by value
+	fmt.Println(c.Age)
```
fwereade: we need to check that the test files compile ;p

tenets/codelingo/go/pass-struct-by-pointer/example.go
fwereade: …and, I guess, it's not so much about "exported" as it is "accessible" – assuming the code compiles (i.e. the fields are accessible) the problematic cases are still possible

PR #452, opened 2019-09-27
Async writes to pointer

CallanTaylor:


tenets/codelingo/go/writes-to-async-pointer/example.go
```diff
@@ -0,0 +1,62 @@
...
+
+func newThingPointer(name string) *thing {
+	return &thing{name}
+}
+
+func newThing(name string) thing {
+	return thing{name}
+}
+
+func main() {
```
fwereade: is there a reason we usually stick all the issues in a single `main` func? One function per issue would be a bit more readable – or at least breaking out the `t2` case to demonstrate that we don't raise an issue when there are no writes happening.

Also, documentation on the components would be really helpful – I look to the examples first to understand the tenet. I'll add some vague suggestions.

tenets/codelingo/go/writes-to-async-pointer/example.go
```diff
@@ -0,0 +1,62 @@
...
+
+func getName(t thing) {
+	fmt.Println(t.name)
+}
+
+func setNameOnPointer(t *thing) {
+	t.name = "car"
+}
+
+func newThingPointer(name string) *thing {
```
fwereade: I don't think this has any explanatory power over using `&thing{name}` in the source directly

tenets/codelingo/go/writes-to-async-pointer/example.go
```diff
@@ -0,0 +1,62 @@
...
+	return &thing{name}
+}
+
+func newThing(name string) thing {
```
fwereade: ditto newThingPointer

tenets/codelingo/go/writes-to-async-pointer/example.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"fmt"
+)
+
+type thing struct {
+	name string
+}
+
+func getNameFromPointer(t *thing) {
```
fwereade: `// readVulnerableField accesses data which may be vulnerable to concurrent writes`?

tenets/codelingo/go/writes-to-async-pointer/example.go
```diff
@@ -0,0 +1,62 @@
...
+	fmt.Println(t.name)
+}
+
+func getName(t thing) {
```
fwereade: `readSafeField` as above?

tenets/codelingo/go/writes-to-async-pointer/example.go
```diff
@@ -0,0 +1,62 @@
...
+	fmt.Println(t.name)
+}
+
+func setNameOnPointer(t *thing) {
```
fwereade: `writeVulnerableField`?

tenets/codelingo/go/writes-to-async-pointer/example.go
```diff
@@ -0,0 +1,62 @@
...
+	return thing{name}
+}
+
+func main() {
+
+	t := newThingPointer("Pointer")
+
+	getNameFromPointer(t)
+
+	go getNameFromPointer(t)
```
fwereade: this block really makes me think we need a sense of read-sections and write-sections

would we catch `go read(); write()`? 'cos it's fundamentally the same issue

CallanTaylor: Are you referring to the Read and Write methods of the io package?

fwereade: I'm referring to the operations in the test file which conflict if run at the same time, which I'd been calling `readVulnerableField`/`writeVulnerableField` above

tenets/codelingo/go/writes-to-async-pointer/expected.json
```diff
@@ -0,0 +1,14 @@
+[
+  {
+   "Comment": "Writing to a reference passed to an asynchronous function can lead to unexpected behaviour.",
```
fwereade: is it possible to include a reference to (one of) the conflicting read(s)?

tenets/codelingo/go/writes-to-async-pointer/example.go
```diff
@@ -0,0 +1,82 @@
...
+
+type thing struct {
+	name string
+}
+
+func readFromPointer(t *thing) {
+	fmt.Println(t.name)
+}
+
+func safeRead(t thing) {
```
fwereade: maybe `readFromCopy`?

tenets/codelingo/go/writes-to-async-pointer/example.go
```diff
@@ -0,0 +1,142 @@
...
+		t.name = "bus"
+	}(t)
+}
+
+// Concurrent reads and writes to a field of a pointer to an instance of thing. Issue
+func concurrentPointerReadWrites() {
+
+	t := &thing{"Pointer"}
+
+	readVulnerableField(t)
```
fwereade: similarly, this should never contribute to one of these issues – it's guaranteed to complete before any of the goroutines kick off

tenets/codelingo/go/writes-to-async-pointer/codelingo.yaml
```diff
@@ -0,0 +1,58 @@
...
+                    go.names:
+                      go.ident:
+                        name as pointer
+                    go.star_expr
+                  go.assign_stmt(depth = any):
+                    go.lhs(depth = any):
+                      go.ident(depth = any):
+                        name == pointer
+            go.call_expr(depth = any):
+              go.func_lit:
```
fwereade: this is not a comment on this code so much as confirmation we need more sophisticated callgraphs – or perhaps something else

to paraphrase the ideal shape I have in my mind, it's more or less:

    some_memory
    goroutine:
      call_some_function(passing=reference_to(some_memory)):
        access(some_memory)
    anything_else_leading_us_to:
      access(some_memory)

…but not actually triggering when _all_ the accesses are reads

tenets/codelingo/go/writes-to-async-pointer/example.go
```diff
@@ -0,0 +1,142 @@
...
+
+	t := &thing{"Pointer"}
+
+	writeVulnerableField(t)
+
+	go writeVulnerableField(t)
+
+	go func(t *thing) {
+		t.name = "bus"
+	}(t)
```
fwereade: also, would this trigger if we just had

    t := &thing{"x"}
    go t.readFromPointer()
    t.name = "y"

or

    t := &thing{"x"}
    go t.writeToPointer("y")
    fmt.Println(t.name)

?

PR #454, opened 2019-10-01
Add Tenet and example

CallanTaylor:


tenets/codelingo/go/naked-chan-access/codelingo.yaml
```diff
@@ -0,0 +1,16 @@
...
+  - name: naked-chan-access
+    actions:
+      codelingo/review:
+        comment: Reading from a channel without wrapping it in a select statement is unsafe. 
+    query: |
+      import codelingo/ast/go
+
+      go.func_decl(depth = any):
+        exclude:
+          go.select_stmt(depth = any):
```
fwereade: might be better to use `comm_clause` here? *might* work better to catch naked channel usage inside the branches

tenets/codelingo/go/naked-chan-access/example.go
```diff
@@ -0,0 +1,38 @@
+package main
+
+import (
+	"time"
+	"fmt"
+)
+
+func inc(i int, c chan int) {
+    c <- i + 1
```
fwereade: this should be an issue too, right?

tenets/codelingo/go/naked-chan-access/example.go
```diff
@@ -0,0 +1,38 @@
...
+		fmt.Println(i)
+	case <- time.After(1 * time.Second):
+		fmt.Println("Timed out")
+	}
+
+	go inc(3, c)
+
+	select {
+	case x = <- c: // Safe
+		fmt.Println(x)
```
fwereade: would we find naked channel use in one of these blocks?

tenets/codelingo/go/naked-chan-access/example.go
CallanTaylor: I had thought that sending to a channel was fine to do outside of a select loop. Is that an anti pattern too do you think?

fwereade: well, it has some nuance, and the risk gets worse the more widely exposed the channel is – my thoughts are mostly in this:

https://github.com/juju/juju/wiki/Boring-Techniques#channels-are-awesome-but-deadly

fwereade: essentially, though – any send or receive that is not in a select is a problem; any range over a channel is also a problem. In both cases, your collaborator at the other end of the channel has the power to block you forever; if you want a robust system, you cannot afford to let that power out of your direct sight and control, or soon enough someone else will "fix a bug" in that other component and break the system.

CallanTaylor:
Closing PR as the Idea has developed into two separate Tenets (one for channel sends and one for receives) which are currently blocked

PR #462, opened 2019-10-09
Shared count datarace

CallanTaylor:
This tenet catches a bug described [here](https://songlh.github.io/paper/go-study.pdf) as Figure 8. It is caused when a loop creates a new go-routine at each iteration and the go-routines created use the iterator of the loop internally. Instead the go-routines should use a copy of the iterator created at the same time as the go-routine by supplying it as an argument


rstorr:
we should catch for ranges as well - add that to example.go

PR #466, opened 2019-11-01
Add tenet and update lingo bundle plus registry

CallanTaylor:


tenets/codelingo/go/only-use-must-in-init/codelingo.yaml
```diff
@@ -0,0 +1,27 @@
+funcs:
+  - name: isNotInit
+    type: asserter
+    body: |
+      function (name) {
+        return !(name == "init")
```
Daanikus:
```suggestion
        return name !== "init"
```

tenets/codelingo/go/only-use-must-in-init/codelingo.yaml
```diff
@@ -0,0 +1,27 @@
...
+      }
+tenets:
+  - name: only-use-must-in-init
+    actions:
+      codelingo/review:
+        comment: Funcitons that begin with Must should only be used in init().
```
Daanikus:
```suggestion
        comment: Functions that begin with Must should only be used in init() or to initialise globals.
```

tenets/codelingo/go/only-use-must-in-init/example.go
```diff
@@ -0,0 +1,44 @@
...
+	bar()
+	init()
+
+}
+
+func foo() {
+
+	var green, _ = regexp.Compile("a")
+
+	yelllow, _ := regexp.Compile("a")
```
Daanikus: s/yelllow/yellow/g

tenets/codelingo/go/only-use-must-in-init/codelingo.yaml
```diff
@@ -0,0 +1,27 @@
+funcs:
```
Daanikus: I feel like this tenet may benefit from a small doc section explaining why we enforcing this. Mention unexpected panics at runtime.

PR #468, opened 2019-11-05
Add ability to catch nested uses of Must and allow Must in main

CallanTaylor:


tenets/codelingo/go/avoid-delayed-must-calls/codelingo.yaml
```diff
@@ -20,8 +20,8 @@ tenets:
         @review comment
         go.assign_stmt(depth = any):
           go.rhs:
-            go.call_expr:
-              go.selector_expr:
+            go.call_expr(depth = any):
+              go.selector_expr(depth = any):
```
Daanikus: I think we can drop the depth = any from the selector expr and also remove everything above call_expr. I can see the assign_stmt preventing capture of things like this:
```
if !regexp.MustCompile("foo").MatchString("bar") {
	fmt.Println("did not match")
}
```

as the result is never assigned to anything.

tenets/codelingo/go/avoid-delayed-must-calls/codelingo.yaml
```diff
@@ -20,8 +20,8 @@ tenets:
...
                 go.ident:
                   name as funcCall
                   regex(/^Must[A-Z]/, funcCall)
```
Daanikus:
```suggestion
                  regex(/^Must([A-Z]|$)/, funcCall)
```

tenets/codelingo/go/avoid-delayed-must-calls/example.go
```diff
@@ -20,7 +21,7 @@ func foo() {
 
 	red := regexp.MustCompile("a") // Issue
 
-	fmt.Println(green + yellow + red)
+	fmt.Println("%s %s %S \n", green.String(), yellow.String(), red.String())
```
Daanikus:
```suggestion
	fmt.Printf("%s %s %s\n", green.String(), yellow.String(), red.String())
```

tenets/codelingo/go/avoid-delayed-must-calls/example.go
```diff
@@ -30,15 +31,17 @@ func bar() {
 
 	red := regexp.MustCompile("a") // Issue
 
-	fmt.Println(green + yellow + red)
+	indigo := regexp.MustCompile("a").Match([]byte(`seafood`)) // Issue
+
+	fmt.Println("%s %s %s %s \n", green.String(), yellow.String(), red.String(), indigo)
```
Daanikus:
```suggestion
	fmt.Printf("%s %s %s %s \n", green.String(), yellow.String(), red.String(), indigo)
```

Daanikus: Is indigo not a bool here? Does this file compile?

tenets/codelingo/go/avoid-delayed-must-calls/example.go
```diff
@@ -30,15 +31,17 @@ func bar() {
...
 func init() {
 	var green, _ = regexp.Compile("a")
 
 	yellow, _ := regexp.Compile("a")
 
-	red := regexp.MustCompile("a")
+	red := regexp.MustCompile("a") // Acceptable
 
-	fmt.Println(green + yellow + red)
+	fmt.Println(green.String() + yellow.String() + red.String())
```
Daanikus: Should change to printf style for consistency

tenets/codelingo/go/avoid-delayed-must-calls/codelingo.yaml
```diff
@@ -3,13 +3,13 @@ funcs:
     type: asserter
     body: |
       function (name) {
-        return name !== "init"
+        return name !== "init" && name !== "main"
```
Daanikus: Maybe rename this func to reflect the inclusion of main. Not a blocking change.

tenets/codelingo/go/avoid-delayed-must-calls/codelingo.yaml
Daanikus: To catch funcs called Must()

tenets/codelingo/go/avoid-delayed-must-calls/example.go
```diff
@@ -10,7 +10,7 @@ func main() {
 	bar()
 
 	red := regexp.MustCompile("a")
-	fmt.Println(red.String())
+	fmt.Printf(red.String())
```
Daanikus: This can be a Println. Is has no formatting verbs.

PR #473, opened 2019-11-15
Add insecure http connection spec

CallanTaylor:


tenets/codelingo/go/insecure-http-connection/example.go
```diff
@@ -0,0 +1,75 @@
...
+			TLSClientConfig: &tls.Config{ // Issue
+				KeyLogWriter: w,
+				Rand: zeroSource{},
+				InsecureSkipVerify: true,
+			},
+		},
+	}
+	resp, err = clientTwo.Get(serverTwo.URL)
+	if err != nil {
+		log.Fatalf("Faile to get url: %v", err)
```
rstorr: s/Faile/Failed. change in rest of file

tenets/codelingo/go/insecure-http-connection/codelingo.yaml
```diff
@@ -0,0 +1,58 @@
+tenets:
+  - name: insecure-http-connection
+    actions:
+      codelingo/review:
+        comment: Skipping verification exposes the connection to man-in-the-middle attacks and should never be used in production code.
```
rstorr: s/ and should never be used in production code./ .This should be used only for testing.

PR #481, opened 2019-12-05
Add concurrent-use-of-rand

CallanTaylor:


yaserazfar:
What about cases where a sync.Mutex is used to ensure that the value is not accessed concurrently? Such as,
`	go func() {
		mu.Lock()
		foo := source.Int63()
		mu.Unlock()
		fmt.Println(foo)
	}()`
Found this from: https://stackoverflow.com/questions/48958886/how-to-create-a-thread-safe-rand-source

PR #486, opened 2020-01-21
Drop public private

fwereade:
golang public/private properties are wrong in many ways:

* strings not bools
* badly named
* like 8% of all our edges in the DB despite being just "is the first letter a capital"

This PR drops all usage of public/private from our specs, in preparation for dropping support for them in platform.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 vendor/
 .DS_Store
 notes.txt
+*.swp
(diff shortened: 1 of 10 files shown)


fwereade:
> cannot build queries: failed to parse codelingo.yaml file `tenets/codelingo/cockroachdb/avoid-bool-params/codelingo.yaml`: line 1:1 - missing import block

…does this always happen? and we've never done anything about it?

PR #490, opened 2020-02-19
Create registry consistency checker.

BlakeMScurr:
Make sure every rule in the registry actually exists in the repo.

The reverse is not true yet, since it probably isn't causing any dashboard problems. However, this utility does tell us which rules aren't used, and it wouldn't be that hard to make it generate the registry file.

### DIFF
--- a/util/registry/main.go
+++ b/util/registry/main.go
@@ -1,0 +1,102 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+
+	"gopkg.in/yaml.v1"
+)
+
+func main() {
+	actualRules := getActualRules()
+	statedRules := getStatedRules()
+	missingActualRules := []string{}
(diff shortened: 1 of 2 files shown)


util/registry/main.go
```diff
@@ -0,0 +1,102 @@
...
+	rules := map[string]bool{}
+	gopath := os.Getenv("GOPATH")
+	registryPath := path.Join(gopath, "src/github.com/codelingo/codelingo/registry/tenets.yaml")
+
+	registryContents, err := ioutil.ReadFile(registryPath)
+	if err != nil {
+		panic(err.Error())
+	}
+
+	r := map[interface{}]map[interface{}]map[interface{}]map[interface{}]interface{}{}
```
Daanikus: Fun times :smile:

BlakeMScurr: Haha, yup - now that's what I call code quality 😆

PR #499, opened 2020-04-13
surprising-for-direction rule

fwereade:
Equivalent to https://lgtm.com/query/rule:1510348425987/lang:go/


tenets/codelingo/go/surprising-for-direction/codelingo.yaml
```diff
@@ -0,0 +1,39 @@
+tenets:
```
rstorr, line 1: can we not use 'rules' yet ?

PR #500, opened 2020-04-20
check for build directive in tparallel rule

Daanikus:


Daanikus:
I think so. That's what Will recommended

rstorr:
can we test it out before merging ? maybe search through the code to see that wherever theres a `.*_integration_test.go` file it contains `// +build integration`

